PR #3, opened 2018-11-06
fix(package): Make sure the package is public for npm for publishing

schenedx:
@reillz10 and @iloveagent57 Please review.
We need this flag to make sure the package is not private for npm so we can publish to it.


fredsmith:
LGTM

PR #4, opened 2018-11-09
Hook up redux to get data for component

reillz10:


src/data/reducers/grades.js
```diff
@@ -0,0 +1,1040 @@
...
+      return {
+        ...state,
+        startedFetching: true,
+        finishedFetching: false,
+      };
+    case ERROR_FETCHING_GRADES:
+      return {
+        ...state,
+        finishedFetching: true,
+        errorFetching: true,
```
schenedx: You might need a `startedFetching: false` statement here as well

src/data/actions/grades.js
```diff
@@ -21,7 +21,22 @@ const fetchGrades = courseId => (
         dispatch(finishedFetchingGrades());
       })
       .catch((error) => {
-        dispatch(errorFetchingGrades())
+        dispatch(errorFetchingGrades());
+      });
+  }
+);
+
+const fetchUserGrades = (courseId, searchText) => (
```
schenedx: I think this action can be `fetchMatchingUserGrades`

reillz10: I agree, that is much better. I didn't know what to name it. Thanks!

PR #15, opened 2018-11-14
Rir/fix category filter

reillz10:


src/data/actions/grades.js
```diff
@@ -32,15 +35,149 @@ const gradeUpdateFailure = error => ({
...
+        key: s.label,
+        columnSortable: false,
+        onSort: (direction) => { this.sortNumerically(s.label, direction); },
+      }));
+
+    return results.concat(assignmentHeadings);
+  },
+};
+
+const filterColumns = (filterType, exampleUser) => ({
```
schenedx: Why did you choose to put these column filter actions in the `grades.js` file? What if we put these logic in a new file?

src/data/actions/grades.js
```diff
@@ -32,15 +35,149 @@ const gradeUpdateFailure = error => ({
...
+    }
+    if (gradeRowA[colKey] > gradeRowB[colKey]) {
+      return -1;
+    }
+    return 0;
+  }
+
+  this.setState({ grades: [...this.state.grades].sort(direction === 'desc' ? sortNumDesc : sortNumAsc) });
+};
+const headingMapper = {
```
schenedx: Move this into a util module so it an be shared by both the grade.js and filter.js

PR #19, opened 2018-11-16
Upgrade to latest frontend-auth.

douglashall:


douglashall:
@iloveagent57 @schenedx @reillz10 This should get the CSRF token auto-retrieval working and improve the login experience by redirecting to the login page before rendering the app if the user is not already authenticated.

schenedx:
@douglashall Does this change solve the issue with POST when we trying to override grades? Also, the auth don't work on `cohort` GET api call. I don't know if that's related.

reillz10:
Handled this with PR #23 so we didn't have to resolve package-lock.json conflicts by hand.

douglashall:
@reillz10 Thanks!

PR #20, opened 2018-11-20
Fix column sorting in the grade table

reillz10:


src/data/actions/utils.js
```diff
@@ -25,39 +26,65 @@ const sortAlphaAsc = (gradeRowA, gradeRowB) => {
 
 const sortNumerically = (colKey, direction) => {
   function sortNumAsc(gradeRowA, gradeRowB) {
-    if (gradeRowA[colKey] < gradeRowB[colKey]) {
+    let a;
```
schenedx: Can you make sure you name the variables what they are? I don't like the meaningless variable names unless they are absolutely meaningless

reillz10: its a common pattern for sorting. We don't actually know what a or b is. It all depends on what is passed in

reillz10: I can call it `valueAtGradeRowA` but I feel like that is just as arbitrary as `a`

src/data/actions/utils.js
```diff
@@ -25,39 +26,65 @@ const sortAlphaAsc = (gradeRowA, gradeRowB) => {
...
-    if (gradeRowA[colKey] > gradeRowB[colKey]) {
+    if (a > b) {
       return 1;
     }
     return 0;
   }
 
   function sortNumDesc(gradeRowA, gradeRowB) {
-    if (gradeRowA[colKey] < gradeRowB[colKey]) {
+    let a;
```
iloveagent57: Everything from here to the `if (a < b)` statement is exactly the same as the lines above.  It's probably worth extracting this to a shared function.

src/data/actions/utils.js
```diff
@@ -25,39 +26,65 @@ const sortAlphaAsc = (gradeRowA, gradeRowB) => {
...
+    }
+
+    if (a < b) {
       return -1;
     }
-    if (gradeRowA[colKey] > gradeRowB[colKey]) {
+    if (a > b) {
       return 1;
     }
     return 0;
```
iloveagent57: Can't you just return `a - b` without checkout for equality?

src/data/actions/utils.js
```diff
@@ -25,39 +26,65 @@ const sortAlphaAsc = (gradeRowA, gradeRowB) => {
...
+    }
+
+    if (a < b) {
       return 1;
     }
-    if (gradeRowA[colKey] > gradeRowB[colKey]) {
+    if (a > b) {
       return -1;
     }
     return 0;
```
iloveagent57: ...and return `b - a` here?

src/data/actions/utils.js
reillz10: Yes I can for the number based sorts.

PR #26, opened 2018-11-29
Add documentation on platform configuration for local development.

iloveagent57:
Just some notes about things you need to do in the edx-platform Django admin in order to use the gradebook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,18 @@ to point to your local gradebook from the instructor dashboard by putting this s
 WRITABLE_GRADEBOOK_URL = 'http://localhost:1991'
 ```
 
+There are also several edx-platform waffle and feature flags you'll have to enable from the Django admin:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,15 @@ to point to your local gradebook from the instructor dashboard by putting this s
 WRITABLE_GRADEBOOK_URL = 'http://localhost:1991'
 ```
 
+There are also several edx-platform waffle and feature flags you'll have to enable from the Django admin:
+
+1. Grades > Persistent grades enabled flag.  Add this flag if it doesn't existing, click the checkboxes to enable
+make it "enabled" and "enabled for all courses".
+2. Waffle > Switches.  Add the "grades.assume_zero_grade_if_absent" switch and make it active.
+3. Waffle_utils > Waffle flag course overrides.  You want to activate this flag for any course
```
robrap: Note: someone could add the waffle flag directly to get this for all courses, or add as course waffle flags if they want to do this per course.

PR #27, opened 2018-11-29
Make it so the assignment time column filter is dynamic and api driven

reillz10:


src/components/Gradebook/index.jsx
```diff
@@ -260,6 +283,16 @@ export default class Gradebook extends React.Component {
                     </label>
                   </span>
                 </div>
+                { this.props.assignmnetTypes.length > 0 &&
```
schenedx: With this drop down, can you please remove the `category` radio button elements on top?

src/components/Gradebook/index.jsx
```diff
@@ -260,6 +283,16 @@ export default class Gradebook extends React.Component {
...
+                  <div className="student-filters">
+                    <InputSelect
```
schenedx: Can you provide a label to this drop down? I think you can just call it `Assignment Types`, just like how Studio does it.

src/components/Gradebook/index.jsx
```diff
@@ -74,6 +75,15 @@ export default class Gradebook extends React.Component {
     return `?${queryString.stringify(parsed)}`;
   };
 
+  mapAssignmentTypeEntries = (entries) => {
+    const mapped = entries.map(entry => ({
+      id: entry,
+      label: entry,
+    }));
+    mapped.unshift({ id: 0, label: 'All-Assignments' });
```
schenedx: If you had the label besides the overall dropdown, you can have this entry to be `All`. That should be clear enough

src/components/Gradebook/index.jsx
```diff
@@ -122,6 +136,15 @@ export default class Gradebook extends React.Component {
     this.props.history.push(updatedQueryStrings);
   };
 
+  mapSelectedAssignmentTypeEntry = (entry) => {
+    const selectedAssignmentTypeEntry = this.props.assignmentTypes
+      .find(x => x.id === parseInt(entry, 10));
+    if (selectedAssignmentTypeEntry) {
+      return selectedAssignmentTypeEntry.name;
+    }
+    return 'All-Assignments';
```
schenedx: Again, I think we should just change this label to `All`

src/data/actions/grades.js
```diff
@@ -72,7 +73,7 @@ const fetchGrades = (courseId, cohort, track, showSuccess) => (
           data.results.sort(sortAlphaAsc),
           cohort,
           track,
-          headingMapper.all(dispatch, data.results[0]),
+          headingMapper('All-Assignments')(dispatch, data.results[0]),
```
schenedx: Maybe move this label text to a single const variable? Please update this to `All`

src/data/apiClient.js
```diff
@@ -10,6 +10,7 @@ const apiClient = getAuthenticatedAPIClient({
   refreshAccessTokenEndpoint: configuration.REFRESH_ACCESS_TOKEN_ENDPOINT,
   accessTokenCookieName: configuration.ACCESS_TOKEN_COOKIE_NAME,
   csrfCookieName: configuration.CSRF_COOKIE_NAME,
+  studioBaseUrl: configuration.STUDIO_BASE_URL,
```
schenedx: This is no longer needed. Please remove.

src/components/Gradebook/index.jsx
reillz10: yes! done!

src/components/Gradebook/index.jsx
reillz10: 👍

PR #30, opened 2018-11-30
fix(filter) filter should remain active after we edit a grade

jansenk:
after editing a grade, rather than doing a blank lookup we should use the current filter

https://openedx.atlassian.net/browse/EDUCATOR-3769

### DIFF
--- a/src/components/Gradebook/index.jsx
+++ b/src/components/Gradebook/index.jsx
@@ -59,7 +59,11 @@ export default class Gradebook extends React.Component {
           earned_graded_override: this.state.updateVal,
         },
       },
-    ]);
+    ],
+    this.state.filterValue,
+    this.props.selectedCohort,
+    this.props.selectedTrack,
+    );
 
     this.setState({
       modalModel: [{}],
--- a/src/data/actions/grades.js
+++ b/src/data/actions/grades.js
@@ -132,14 +132,14 @@ const fetchPrevNextGrades = (endpoint, cohort, track) => (
 );
 
 
-const updateGrades = (courseId, updateData) => (
+const updateGrades = (courseId, updateData, searchText, cohort, track) => (
   (dispatch) => {
     dispatch(gradeUpdateRequest());
     return LmsApiService.updateGradebookData(courseId, updateData)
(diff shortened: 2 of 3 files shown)


src/components/Gradebook/index.jsx
```diff
@@ -59,7 +59,11 @@ export default class Gradebook extends React.Component {
           earned_graded_override: this.state.updateVal,
         },
       },
-    ]);
+    ],
+    this.state.filterValue,
```
schenedx: The indentation here doesn't look right. Please correct.

src/data/actions/grades.js
```diff
@@ -132,14 +132,14 @@ const fetchPrevNextGrades = (endpoint, cohort, track) => (
...
-const updateGrades = (courseId, updateData) => (
+const updateGrades = (courseId, updateData, searchText, cohort, track) => (
   (dispatch) => {
     dispatch(gradeUpdateRequest());
     return LmsApiService.updateGradebookData(courseId, updateData)
       .then(response => response.data)
       .then((data) => {
         dispatch(gradeUpdateSuccess(data));
-        dispatch(fetchGrades(courseId, null, null, true));
+        dispatch(fetchMatchingUserGrades(courseId, searchText, cohort, track));
```
schenedx: What if we did not have any search texts in the fields, would this still work?
Basically, if the grade adjustment operated on top of default all view, would this change cause breakage?

jansenk: it shouldn't break anything. it calls LmsApiService.fetchGradebookData which handles null arguments correctly

schenedx: Could you please test that case to make sure?

jansenk: seems to be fine!

src/components/Gradebook/index.jsx
jansenk: I'm not sure what you mean. What should I change here?

schenedx: So, the ] should be combined with the } above. Then you can have the values indented a "tab" in.
Anyways. Not a big deal.

PR #31, opened 2018-12-03
Show a spinner when waiting for the grades call to come back

reillz10:


src/components/Gradebook/gradebook.scss
```diff
@@ -1,3 +1,16 @@
+.rick{
```
schenedx: Trying to sneak this in here? come on....

PR #33, opened 2018-12-03
fix(UI) box should appear after editing grade

jansenk:


src/data/actions/grades.js
```diff
@@ -87,7 +87,7 @@ const fetchGrades = (courseId, cohort, track, showSuccess) => (
   }
 );
 
-const fetchMatchingUserGrades = (courseId, searchText, cohort, track) => (
+const fetchMatchingUserGrades = (courseId, searchText, cohort, track, showSuccess) => (
```
schenedx: I feel like `showSuccess` is not a good name for the variable. 
Can the name be something like `showUpdateSuccessfulBanner`?

src/data/actions/grades.js
```diff
@@ -102,6 +102,7 @@ const fetchMatchingUserGrades = (courseId, searchText, cohort, track) => (
           data.next,
         ));
         dispatch(finishedFetchingGrades());
+        dispatch(updateBanner(!!showSuccess));
```
reillz10: no need for `!!` What that does is cast a truthy/falsey to a boolean. Since we already started with a boolean there is no need for it.

jansenk:
that's not what i wanted that commit message to be x(

reillz10:
@jansenk you should squash your commits down to one message for this work

PR #36, opened 2018-12-04
fix(UI) rounded percentages to two decimal places

jansenk:


src/components/Gradebook/index.jsx
```diff
@@ -185,11 +185,11 @@ export default class Gradebook extends React.Component {
               className="btn btn-header link-style"
               onClick={() => this.setNewModalState(entry, subsection)}
             >
-              {subsection.percent * 100}%
+              {parseFloat((subsection.percent * 100).toFixed(2))}%
             </button>);
           return acc;
         }, {});
-      const totals = { total: `${entry.percent * 100}%` };
+      const totals = { total: `${parseFloat((entry.percent * 100).toFixed(2))}%` };
```
schenedx: Can you please move the rounding number to a const in this class? Also, is 2 decimal enough? Please check with Deen

jansenk: I checked with him, he said 2 is good

iloveagent57: I think you'd still want to store 2 as a constants somewhere, like `DECIMAL_PRECISION = 2` or something, since we're using it in more than one place.

PR #37, opened 2018-12-04
ARCH-308: Reimplement LMS footer in React in Gradebook.

robrap:
[ARCH-308](https://openedx.atlassian.net/browse/ARCH-308)
Note: See subtasks in JIRA for some outstanding work.

**Description**

Re-implements the LMS footer in React in the Gradebook app.  A future story will make this a reusable component.

Note: This implementation uses different styling from the original.

**Sandbox**
[https://robert.edxdev.org/course-v1:edX+DemoX+Demo_Course](https://robert.edxdev.org/course-v1:edX+DemoX+Demo_Course)

**To do**
- [x] Add reviewers for styling (after confirming CSS Grid ok in IE?)

**FedX Questions**
- Old footer used id footer-edx-v3.  Why?
- Reset CSS file? 
  - Do we care about minor style differences in text from original?

**Implementation Notes**
This footer uses different mark-up and styling implementation from the original:
- It does not use an id like `footer-edx-v3` for the footer.
- It does not use bootstrap's grid system and corresponding markup.


robrap:
@AlasdairSwan: This is still a WIP, including verifying css grid works ok on IE.  Also working on sandbox.

src/components/Gradebook/footer.jsx
```diff
@@ -0,0 +1,89 @@
...
+import {Icon} from '@edx/paragon';
+
+export default class Footer extends React.Component {
+
+  render() {
+    return (
+      <footer role="contentinfo" aria-label="Page Footer"
+              className="d-flex justify-content-center border-top py-3 px-4">
+        <div className="max-width-1180 d-grid">
+          <div className="area-1"><img src="assets/logo-footer.png" /></div>
```
AlasdairSwan: ```img``` elements should have an ```alt``` attribute.

src/components/Gradebook/footer.jsx
```diff
@@ -0,0 +1,89 @@
...
+          <div className="area-2">
+            <h2>edx</h2>
```
AlasdairSwan: Styles should be applied to classes instead of elements.

src/components/Gradebook/footer.jsx
```diff
@@ -0,0 +1,89 @@
...
+              <li><a href="https://www.edx.org/about-us">About</a></li>
+              <li><a href="https://www.edx.org/enterprise">edX for Business</a></li>
+              <li><a href="https://www.edx.org/affiliate-program">Affiliates</a></li>
+              <li><a href="http://open.edx.org">Open edX</a></li>
+              <li><a href="https://www.edx.org/careers">Careers</a></li>
+              <li><a href="https://www.edx.org/news-announcements">News</a></li>
+            </ul>
+          </div>
+          <div className="area-3">
+            <h2>Legal</h2>
```
AlasdairSwan: Styles should be applied to classes instead of elements.

src/components/Gradebook/footer.jsx
```diff
@@ -0,0 +1,89 @@
...
+            <ul className="list-unstyled">
+              <li><a href="https://www.edx.org/edx-terms-service">Terms of Service &amp; Honor Code</a></li>
+              <li><a href="https://www.edx.org/edx-privacy-policy">Privacy Policy</a></li>
+              <li><a href="https://www.edx.org/accessibility">Accessibility Policy</a></li>
+              <li><a href="https://www.edx.org/trademarks">Trademark Policy</a></li>
+              <li><a href="https://www.edx.org/sitemap">Sitemap</a></li>
+            </ul>
+          </div>
+          <div className="area-4">
+            <h2>Connect</h2>
```
AlasdairSwan: Styles should be applied to classes instead of elements.

src/components/Gradebook/footer.jsx
```diff
@@ -0,0 +1,89 @@
...
+              <li><a href="https://itunes.apple.com/us/app/edx/id945480667?mt=8" target="_blank">
+                <img className="max-height-39"
+                     alt="Download the edX mobile app from the Apple App Store"
+                     src="https://prod-edxapp.edx-cdn.org/static/images/app/app_store_badge_135x40.d0558d910630.svg" />
+              </a></li>
+              <li><a href="https://play.google.com/store/apps/details?id=org.edx.mobile" target="_blank">
+                <img className="max-height-39"
+                     alt="Download the edX mobile app from Google Play"
+                     src="https://prod-edxapp.edx-cdn.org/static/images/app/google_play_badge_45.6ea466e328da.png" />
+              </a></li>
```
AlasdairSwan: There is some weird formatting going on in the file, it would be great to add the edX JS linter to this repo for consistency (https://github.com/edx/eslint-config-edx).

src/components/Gradebook/footer.scss
```diff
@@ -0,0 +1,168 @@
...
+
+.max-height-39 {
+  max-height: 39px;
+}
+
+.d-grid {
+  display: grid;
+}
+
+$border-1: 1px solid #e7e7e7;
```
AlasdairSwan: Variables should be used for colours.

src/components/Gradebook/footer.scss
```diff
@@ -0,0 +1,168 @@
...
+
+footer {
+  background-color: #fcfcfc;
```
AlasdairSwan: Variables should be used for colours.

src/components/Gradebook/footer.scss
```diff
@@ -0,0 +1,168 @@
...
+  max-height: 39px;
+}
+
+.d-grid {
+  display: grid;
+}
+
+$border-1: 1px solid #e7e7e7;
+
+footer {
```
AlasdairSwan: Styles should be applied to classes instead of elements.

src/components/Gradebook/footer.scss
```diff
@@ -0,0 +1,168 @@
...
+  background-color: #fcfcfc;
+
+  ul {
```
AlasdairSwan: Styles should be applied to classes instead of elements.

src/components/Gradebook/footer.scss
```diff
@@ -0,0 +1,168 @@
...
+    padding: 0px;
+    margin: 0px;
```
AlasdairSwan: No need for units if the value is ```0```.

src/components/Gradebook/footer.scss
```diff
@@ -0,0 +1,168 @@
...
+    grid-row: 4;
+    border-bottom: $border-1;
+    padding: 1rem 0;
+  }
+
+  .area-5 {
+    grid-column: 1;
+    grid-row: 5;
+    padding: 1rem 0;
+  }
```
AlasdairSwan: Nitpick: this could be a little DRYer is a common ```.area``` class was used as well (given that they share the same ```grid-column``` and ```padding``` values. But that may lead to confusion later when the classes deviate, so I'm also fine with leaving as is.

src/components/Gradebook/footer.jsx
robrap: See similar comment above.

src/components/Gradebook/footer.jsx
robrap: See similar comment above.

src/components/Gradebook/footer.scss
robrap: Done.

src/components/Gradebook/footer.scss
robrap: Done.

src/components/Gradebook/footer.scss
robrap: Done.
FYI: @reillz10: Might want to create a variables file.

src/components/Gradebook/footer.scss
robrap: Moved to jsx.

src/components/Gradebook/footer.scss
robrap: deleted.

src/components/Gradebook/footer.scss
robrap: Going to leave as-is.  Not sure how and if styling may change when we make this a shared component.

robrap:
Only known blocker is the footer logo was broken.

albemarle:
@robrap @douglashall The logo is broken on the AWS sandboxes, but looks fine for me running locally.  (See screenshot.) . Should I just hardcode it with the CDN URL like the other images are?

image

douglashall:
@albemarle See https://github.com/edx/gradebook/blob/master/src/components/Header/index.jsx#L4.

wittjeff:
id footer-edx-v3 is a link target for a **Skip to end of results** link that appears if you tab through the (very very long list) of results on edx.org/course page

wittjeff:
I suggest making the current (end-of-range) copyright date auto-updating.

wittjeff:
Possible TODO (not WCAG-required): make the social media icons High Contrast Mode-safe.

robrap:
@wittjeff: 
1. The id for "skip to" I'd like to wait on until it is used/needed with this componenet.  At that time, we can see if that continues to be the best solution, and maybe we can use a less strange id if this is part of the public interface of the footer.
2. Re: copyright, agreed.  It is part of a follow-up ticket we will work on.
3. We can look at a potential addition for the High Contrast Mode-safe icons as part of the remaining tasks.
Thanks.

PR #47, opened 2018-12-10
fix(UI): round numerator of grade ratios.

iloveagent57:
When selecting "absolute" as the score view, you can get floats in need of rounding, as depicted in the attached image.  This PR rounds the numerator of the absolute total grade.  It also rounds the numerator and denominator of absolute subsection grades, for consistency.
screen shot 2018-12-10 at 2 08 07 pm


iloveagent57:
After fix:
screen shot 2018-12-10 at 2 10 02 pm

jansenk:
Ah, I should have thought of this X(

PR #50, opened 2018-12-11
fix(feat): Prevent editing if course grades are frozen

schenedx:
This is to display the alert to the user that course has the grades frozen. In this case, we don't allow grades editing. See screenshot:

frozen_grades

@edx/educator-neem Please review


jansenk:
I think it should be "The grades for this course **are** now frozen. Editing of grades **is** no longer allowed"

schenedx:
@jansenk Yes. My grammar is incorrect. I updated the wording according to your suggestion now

src/components/Gradebook/index.jsx
```diff
@@ -179,36 +179,44 @@ export default class Gradebook extends React.Component {
   roundGrade = percent => parseFloat(percent.toFixed(DECIMAL_PRECISION));
 
   formatter = {
-    percent: entries => entries.map((entry) => {
+    percent: (entries, isGradesFrozen) => entries.map((entry) => {
```
iloveagent57: Grammar nit: should be `areGradesFrozen`

src/components/Gradebook/index.jsx
```diff
@@ -179,36 +179,44 @@ export default class Gradebook extends React.Component {
...
+              </button>);
+          }
           return acc;
         }, {});
       const totals = { total: `${this.roundGrade(entry.percent * 100)}%` };
       return Object.assign(results, assignments, totals);
     }),
 
-    absolute: entries => entries.map((entry) => {
+    absolute: (entries, isGradesFrozen) => entries.map((entry) => {
```
iloveagent57: Same thing here.

src/components/Gradebook/index.jsx
```diff
@@ -233,6 +241,11 @@ export default class Gradebook extends React.Component {
             </a>
             <h1>Gradebook</h1>
             <h3> {this.props.match.params.courseId}</h3>
+            { this.props.isGradesFrozen &&
+              <div className="alert alert-danger" role="alert" >
+                The grades for this course are now frozen. Editing of grades is no longer allowed!
```
iloveagent57: Grammar nit: is the exclamation point really necessary?

PR #52, opened 2018-12-12
POC Reactstrap Navbar-based Header

douglashall:


douglashall:
@schenedx @reillz10 @iloveagent57 FYI, this is not something you will want to merge, but I did want to point out the use of the `userProfile` reducer from `@edx/frontend-auth` which might be useful for you if you are looking to inject user profile data into your Redux state.

PR #55, opened 2018-12-14
render error message if user is not allowed to view gradebook

jansenk:
EDUCATOR-3796


src/data/services/LmsApiService.js
```diff
@@ -4,6 +4,9 @@ import { configuration } from '../../config';
 class LmsApiService {
   static baseUrl = configuration.LMS_BASE_URL;
   static pageSize = 10
+  static rolesUrl = `${LmsApiService.baseUrl}/api/enrollment/v1/roles/`;
```
schenedx: Please follow existing pattern and move this variable into the `canUserViewGradebook` function

src/data/services/LmsApiService.js
```diff
@@ -4,6 +4,9 @@ import { configuration } from '../../config';
...
+  static allowed_roles = ['staff', 'instructor', 'support'];
```
schenedx: Please move this as well.

src/index.jsx
```diff
@@ -25,6 +26,15 @@ const App = () => (
   </Provider>
 );
 
+const Unauthorized = () => "You are not authorized to access the gradebook for this course.";
+
 if (apiClient.ensurePublicOrAuthencationAndCookies(window.location.pathname)) {
-  ReactDOM.render(<App />, document.getElementById('root'));
+  LmsApiService.canUserViewGradebook(window.location.pathname.substring(1)).then(
```
schenedx: Could you please give an variable to `window.location.pathname.substring(1)`? I don't really know what this substring represent. Also, please handle the case if the `substring(1)` is empty or null

src/index.jsx
```diff
@@ -25,6 +26,15 @@ const App = () => (
   </Provider>
 );
 
+const Unauthorized = () => "You are not authorized to access the gradebook for this course.";
```
schenedx: I recommend we add an action link to this element. Say `Click here` to go back or something.

src/index.jsx
jansenk: this is the course key. it includes the leading "/", which the substring removes.
You're right, I should be checking if it's empty

src/index.jsx
jansenk: Go back where?

src/index.jsx
```diff
@@ -26,10 +29,21 @@ const App = () => (
   </Provider>
 );
 
-const Unauthorized = () => "You are not authorized to access the gradebook for this course.";
+const Unauthorized = () => (
+  <div>
+    You are not authorized to access the gradebook for this course.
+    &nbsp;
+    <a
+      href={`${configuration.LMS_BASE_URL}/courses/${courseId}/instructor`}
```
iloveagent57: What if they don't have access to the instructor dashboard?

jansenk: Then they shouldn't have been going to the gradebook in the first place ;)

jansenk: But actually, I don't know. what would you suggest?

src/index.jsx
```diff
@@ -25,6 +29,26 @@ const App = () => (
   </Provider>
 );
 
+const Unauthorized = () => (
+  <div>
+    You are not authorized to access the gradebook for this course.
```
schenedx: Maybe the message can be improved to say `You are not enrolled or authorized to access the gradebook for this course`?

src/index.jsx
iloveagent57: I dunno, depends on who this message is for.  If it's primarily for course team members that somehow don't have access to this page, then your current message with a link back to the instructor dashboard is probably fine.

jansenk:
screen shot 2018-12-18 at 2 01 36 pm

src/components/Gradebook/index.jsx
```diff
@@ -259,6 +252,11 @@ export default class Gradebook extends React.Component {
                 The grades for this course are now frozen. Editing of grades is no longer allowed.
               </div>
             }
+            { !this.props.canUserViewGradebook &&
+              <div className="alert alert-warning" role="alert" >
+                You are not authorized to view the gradebook for this course. If you have a global role, try enrolling in this course and try again.
```
schenedx: The word `try` is used twice. Please only use that word once, like `If you have a global role, please enroll into this course and try again`

src/containers/GradebookPage/index.jsx
```diff
@@ -28,10 +29,21 @@ const mapStateToProps = state => (
     nextPage: state.grades.nextPage,
     assignmnetTypes: state.assignmentTypes.results,
     areGradesFrozen: state.assignmentTypes.areGradesFrozen,
-    showSpinner: state.grades.showSpinner,
+    showSpinner: isShowSpinner(state),
+    canUserViewGradebook: state.roles.canUserViewGradebook
   }
 );
 
+function isShowSpinner (state) {
```
schenedx: Better function name would be `shouldShowSpinner`

PR #62, opened 2018-12-20
fix(auth) allow global staff to view gradebook

jansenk:


src/data/actions/roles.js
```diff
@@ -17,8 +17,9 @@ const getRoles = (courseId, urlQuery) => (
   (dispatch) => {
     return LmsApiService.fetchUserRoles(courseId)
       .then(response => response.data)
-      .then(roles => {
-        var canUserViewGradebook = roles.some(role => (role.course_id === courseId) && allowed_roles.includes(role.role));
+      .then(response => {
+        var canUserViewGradebook = response.roles.some(role => (role.course_id === courseId) && allowed_roles.includes(role.role));
+        canUserViewGradebook = canUserViewGradebook || response.is_staff;
```
schenedx: Can you move the statement `response.is_staff` ahead of your roles logic 1 line above? This way you can take advantage of short-circuiting.

PR #66, opened 2018-12-21
implement gradebook pagination button feedback

jansenk:
refactor buttons to a pure function component
change labels
disable rather than hide

EDUCATOR-3825

### DIFF
--- a/src/components/Gradebook/index.jsx
+++ b/src/components/Gradebook/index.jsx
@@ -10,6 +10,7 @@ import {
 } from '@edx/paragon';
 import queryString from 'query-string';
 import { configuration } from '../../config';
+import PageButtons from '../PageButtons';
 
 const DECIMAL_PRECISION = 2;
 
@@ -329,21 +330,6 @@ export default class Gradebook extends React.Component {
                   onClear={() => this.props.getUserGrades(this.props.match.params.courseId, this.props.selectedCohort, this.props.selectedTrack)}
                   value={this.state.filterValue}
                 />
-                <div className="d-flex justify-content-end" style={{ marginTop: '20px' }}>
-                  <Button
-                    label="Previous"
-                    buttonType="primary"
-                    style={{ visibility: (!this.props.prevPage ? 'hidden' : 'visible') }}
-                    onClick={() => this.props.getPrevNextGrades(this.props.prevPage, this.props.selectedCohort, this.props.selectedTrack)}
-                  />
-                  <div style={{ width: '10px' }} />
-                  <Button
-                    label="Next"
-                    buttonType="primary"
-                    style={{ visibility: (!this.props.nextPage ? 'hidden' : 'visible') }}
-                    onClick={() => this.props.getPrevNextGrades(this.props.nextPage, this.props.selectedCohort, this.props.selectedTrack)}
-                  />
-                </div>
               </div>
             </div>
             <br />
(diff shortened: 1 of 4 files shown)


src/components/Gradebook/index.jsx
```diff
@@ -357,6 +343,7 @@ export default class Gradebook extends React.Component {
               onClose={() => this.props.updateBanner(false)}
               open={this.props.showSuccess}
             />
+            {PageButtons(this.props)}
```
schenedx: The way data passed to components here can be improved. @reillz10 please help out here.

src/components/PageButtons/index.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import {
+  Button,
+} from '@edx/paragon';
+
+function PageButton(label, props, page) {
```
schenedx: This is not a component in reactjs. We should make sure we use the reactjs component interface. See https://github.com/edx/gradebook/blob/master/src/components/Header/index.jsx#L6 as an example

jansenk: I originally did that, but eslint told me that since this component is stateless it should be a pure function component

reillz10: I'm not sure that the abstraction from `Button` to `PageButton` is actually helpful here yet. I would just put those two buttons as is in the PageButtons component. Then you should destructure props in the function paramater signature. (e.x. `function PageButtons({selectedCohort, prevPage, nextPage, selectedTrack})`)

src/components/PageButtons/PageButtons.test.jsx
```diff
@@ -0,0 +1,29 @@
...
+    prevPage,
+    nextPage,
+    selectedTrack: 't',
+    selectedCohort: 'c',
+    getPrevNextGrades() {},
+  };
+};
+
+describe('PageButtons', () => {
+  it('renders empty links as disabled', () => {
```
schenedx: Maybe we should break this test case into multiple test cases? Also, you are repeating statements in these tests a lot, is there a way to refactor?

src/components/PageButtons/index.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import {
+  Button,
+} from '@edx/paragon';
+
+
+export default function PageButtons({prevPage, nextPage, selectedTrack, selectedCohort, getPrevNextGrades}) {
```
schenedx: Instead of taking the arguments as a dictionary like this, what is the challenge of specify each of the dictionary key as argument item?
How difficult would it be to just do 
`PageButtons(prevPage, nextPage, selectedTrack, selectedCohort, getPrevNextGrades)`?

schenedx: Talked about this offline and apparently, this is the better choice to take in parameters on this component. Let's ignore my comment above

src/components/PageButtons/PageButtons.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+    it('prev null, next not null', () => {
+      assertPageButtonsSnapshot(createInput(null, 'np'));
+    });
+
+    it('prev not null, next null', () => {
+      assertPageButtonsSnapshot(createInput('pp', null));
+    });
+
+    it('prev not null, next not null', () => {
```
schenedx: #NIT it would be clearer for the test cases if this is pulled out into it's own test `describe` function

PR #73, opened 2019-01-07
move project run/setup to devstack

Zacharis278:
~This depends on https://github.com/edx/devstack/pull/379.   Without it there is no way to actually run the project.~
Edit: This is now in master


Makefile
```diff
@@ -1,35 +1,12 @@
-shell: ## run a shell on the cookie-cutter container
-	docker exec -it edx.gradebook /bin/bash
+shell: ## run a shell
```
iloveagent57: Will I be able to do `make gradebook-shell` from inside devstack now?

README.md
```diff
@@ -23,18 +23,18 @@ npm i --save @edx/gradebook
 
 After cloning the repository, run `make up-detached` in the `gradebook` directory - this will build and start the `gradebook` web application in a docker container.
```
iloveagent57: The `up-detached` command was deleted, so this line should be updated.

Zacharis278: good call

Makefile
Zacharis278: Yeah you can get to shell via devstack as well I just left it as a shortcut.  Ultimately, it's unnecessary.  However there's no equivalent of the npm install shortcut in devstack so we probably want to keep that.

Zacharis278:
@iloveagent57 I've made the requested updates and removed all docker references from the makefile

PR #74, opened 2019-01-08
add metadata for openedx releases

Zacharis278:
Followup to [EDUCATOR-3853](https://openedx.atlassian.net/browse/EDUCATOR-3853)

@nedbat mentioned we need this file for openedx since gradebook is a main project installed within devstack.

### DIFF
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -1,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: grbk
+oeps: {}
+owner: edx/educator-neem
+openedx-release: {ref: master}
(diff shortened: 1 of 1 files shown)


openedx.yaml
```diff
@@ -0,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: grbk
+oeps: {}
```
iloveagent57: Since the `oeps` key is optional, I think you can safely eliminate it.

Zacharis278: huh it does say optional, yet the other projects I'm looking at have an empty object.

iloveagent57: Hmm, I guess it doesn't really matter.  You can leave as is.

PR #76, opened 2019-01-15
Add segment library integration with Gradebook to track events

schenedx:
@edx/educator-neem Please review.
This PR mimic what's done in https://github.com/edx/edx-portal/pull/29.
It's not meant to be a complete capture of all the events on gradebook. This is an integration to setup tracking on gradebook

All the events should stream into segment.io after deployment.

### DIFF
--- a/src/data/store.js
+++ b/src/data/store.js
@@ -2,14 +2,48 @@ import { applyMiddleware, createStore } from 'redux';
 import thunkMiddleware from 'redux-thunk';
 import { composeWithDevTools } from 'redux-devtools-extension/logOnlyInProduction';
 import { createLogger } from 'redux-logger';
+import { createMiddleware } from 'redux-beacon';
+import Segment, { trackEvent, trackPageView } from '@redux-beacon/segment';
+import { GOT_ROLES } from './constants/actionTypes/roles';
+import { GOT_GRADES, GRADE_UPDATE_SUCCESS, GRADE_UPDATE_FAILURE } from './constants/actionTypes/grades';
 
 import reducers from './reducers';
 
 const loggerMiddleware = createLogger();
 
+const eventsMap = {
+  [GOT_ROLES]: trackPageView(action => ({
+    page: action.courseId,
+  })),
+  [GOT_GRADES]: trackEvent(action => ({
+    name: 'Grades displayed or paginated',
+    properties: {
+      track: action.track,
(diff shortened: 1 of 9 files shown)


src/segment.js
```diff
@@ -0,0 +1,85 @@
...
+    return function () {
+      const args = Array.prototype.slice.call(arguments);
+      args.unshift(method);
+      analytics.push(args);
+      return analytics;
+    };
+  };
+
+  // For each of our methods, generate a queueing stub.
+  for (let i = 0; i < analytics.methods.length; i++) {
```
iloveagent57: Isn't this just `analytics.methods.forEach(methodName => analytics[methodName] = analytics.factory(methodName);)`?

src/segment.js
```diff
@@ -0,0 +1,85 @@
...
+
+  // Define a factory to create stubs. These are placeholders
+  // for methods in Analytics.js so that you never have to wait
+  // for it to load to actually record data. The `method` is
+  // stored as the first argument, so we can replay the data.
+  analytics.factory = function (method) {
+    return function () {
+      const args = Array.prototype.slice.call(arguments);
+      args.unshift(method);
+      analytics.push(args);
```
iloveagent57: I don't understand this.

src/data/store.js
```diff
@@ -2,14 +2,42 @@ import { applyMiddleware, createStore } from 'redux';
...
+  [GOT_GRADES]: trackEvent(action => ({
+    name: 'Grades displayed or paginated',
+    properties: {
+      track: action.track,
+      cohort: action.cohort,
+      prev: action.prev,
+      next: action.next,
+    },
+  })),
+  [GRADE_UPDATE_SUCCESS]: trackEvent(action => ({
```
iloveagent57: Is it worth tracking the failures, too?

schenedx: Good point.

src/segment.js
schenedx: This is not my code. This whole file comes from segment.js. I rather not making changes here.

iloveagent57: Yeah, good point.  I am curious in general about this type of pattern, but I can look it up on the internet.

src/data/store.js
schenedx: @iloveagent57 This is now eventing on both success and failure cases.

PR #85, opened 2019-01-24
EDUCATOR-3936 Increase users per page from 10 to 25

kdmccormick:
@edx/educator-neem


iloveagent57:
There's also a unit test where you need to change "10" to "25" :)

PR #87, opened 2019-01-25
Remove remove all unneeded sorting

reillz10:


Zacharis278:
@kdmccormick humm yeah it's likely the change removing the default assignment type fallback.  I can probably take a look since I just worked on that if we need to get this merged before @reillz10 gets back.

PR #94, opened 2019-05-28
Refactored unnecessarily verbose mapDispatchToProps

matthugs:
Also split up paramaterized/generic "updateBanner" action to two
separate actions which are dispatched only when necessary and added an
npm run script for Jest's watch mode

I mostly made these changes to get my bearings a little bit and to make the repo easier to work with. I'm particularly interested in making it easier to split off a new container for the grade change modal (since I'll be adding a new modal for a different purpose soon).

if the mapDispatchToProps in particular looks like witchcraft to you, I'd encourage you to check out https://react-redux.js.org/using-react-redux/connect-mapdispatch#defining-mapdispatchtoprops-as-an-object


matthugs:
Got an OTS ok from @reillz10 https://edx-internal.slack.com/archives/CD9K8CKH7/p1559837150008700?thread_ts=1559836586.008400&cid=CD9K8CKH7 on this, but I'm rolling this into #95 rather than pushing it forward separately.

PR #97, opened 2019-06-13
mroytman/add email student key

MichaelRoytman:


src/components/Gradebook/gradebook.scss
```diff
@@ -41,24 +45,37 @@
 .gbook { 
     overflow-x: scroll;
 
+    .grade-button {
+        text-decoration: underline;
```
MichaelRoytman: UX/UI request: "For the grades themselves, suggest underlining the text. Affordance around this interaction will help with how intuitive it feels."

src/components/Gradebook/gradebook.scss
```diff
@@ -41,24 +45,37 @@
...
+
+    .student_key {
+        font-size: 14px;
+    }
+
     .table {
         padding-left: 244px;
+
+        th {
+            vertical-align: top;
```
MichaelRoytman: UX/UI request

"Table head row: suggest vertical alignment top"
"Table head row: suggest slightly smaller for font size, 14px. I realize this is smaller than our standard 16px, but I think it's necessary given how much text we are squeezing into each cell"

src/components/Gradebook/gradebook.scss
```diff
@@ -27,6 +27,10 @@
     @media only screen and (min-width: 1200px) {
         width: 1024px;
     }
+
+    .search-help-text {
+        margin-left: 20px;
```
MichaelRoytman: UX/UI request: search help text should be aligned with the label in the SearchField

src/components/Gradebook/index.jsx
```diff
@@ -334,7 +365,7 @@ export default class Gradebook extends React.Component {
                       this.props.selectedAssignmentType,
                     )
                   }
-                  inputLabel="Search Username:"
+                  inputLabel="Search for a learner"
```
MichaelRoytman: Note: Christine said we should remove the colon here.

src/components/Gradebook/index.jsx
```diff
@@ -346,6 +377,7 @@ export default class Gradebook extends React.Component {
                   }
                   value={this.state.filterValue}
                 />
+                <small className="form-text text-muted search-help-text">Search by username, email, or student key</small>
```
MichaelRoytman: Using the same element/classname combination as the help text in the language section of the new account settings page.

src/components/Gradebook/index.jsx
```diff
@@ -358,7 +390,8 @@ export default class Gradebook extends React.Component {
             {PageButtons(this.props)}
             <div className="gbook">
               <Table
-                columns={this.props.headings}
+                className={['table-striped']}
```
MichaelRoytman: UX/UI request: "Username rows: Starting with the first username row, background color of #f5f5f5 for every other row. More delineations between rows helps clean up the table and makes scanning the middle of the table a little easier"

src/data/actions/utils.js
```diff
@@ -11,54 +11,27 @@ const sortAlphaAsc = (gradeRowA, gradeRowB) => {
 };
 
 const headingMapper = (filterKey) => {
-  function all(entry) {
+  const filters = {
```
MichaelRoytman: I have chosen to remove the label from Redux, as I didn't want to tie the label that is displayed for each column into Redux, particularly since the Username column label is now an element and not just a string.

Since we use this state to determine what filters have been applied to the headings, I have changed it to an array of keys.

MichaelRoytman: I have also done some refactoring to make this function more DRY.

src/components/Gradebook/index.jsx
```diff
@@ -220,13 +231,33 @@ export default class Gradebook extends React.Component {
...
+        <div>
+          <div>Username</div>
+          <div className="font-weight-normal student_key">Student Key</div>
+        </div>
+      );
+
+      headings = headings.map(heading => ({ label: heading, key: heading }));
+
+      // replace username heading label to include additional user data
+      headings[0].label = userInformationHeadingLabel;
```
MichaelRoytman: I'd like to think of a better way to do this, as this assumes that the username column will always be at this index but have not had time to look into it further.

src/components/Gradebook/gradebook.scss
```diff
@@ -41,24 +45,35 @@
...
+            font-size: 14px;
+        }
     }
 
     .table tr th:first-child {
         position: absolute;
         width: 160px;
         height:50px;
         display: block;
-        background-color: #fff;
```
MichaelRoytman: This was mucking with the striped rows.

src/components/Gradebook/index.jsx
```diff
@@ -220,13 +231,33 @@ export default class Gradebook extends React.Component {
...
   lmsInstructorDashboardUrl = courseId => `${configuration.LMS_BASE_URL}/courses/${courseId}/instructor`;
 
+  formatHeadings = () => {
+    let headings = [...this.props.headings];
+
+    if (headings.length > 0) {
+      const userInformationHeadingLabel = (
+        <div>
+          <div>Username</div>
+          <div className="font-weight-normal student_key">Student Key</div>
```
MichaelRoytman: TODO: The student key heading should not be display if there are no student keys in the data.

src/components/Gradebook/index.jsx
MichaelRoytman: TODO: If email is not being displayed in the gradebook (e.g. for a non-Master's course), we may want to consider removing email from these instructions to avoid confusing the user.

MichaelRoytman:
TODOs:
* UX/UI request: "Username rows: It looks like the column of usernames is aligned top, but it should be middle to match the row of grades" This is true of the Email and Total columns as well.
* UX/UI request: "Username rows: If we can control the height of rows, suggest height of ~15-20px, compressing the height a little I think will help with viewing long lists"

* The Email column should be shown conditionally . From the [epic](https://openedx.atlassian.net/browse/EDUCATOR-4281), it seems we want "V1: limited to Master's track only". I'm not sure what this mean. Show emails for all learners for a course with a Master's track? Only show emails for learners in a Master's track for a course with a Master's track? Something else? This should be clarified. 
* The "Student Key" label in the row header should be dynamic. If there is no external user key in the current data, do not show this label.
* The styling of the button interferes with the striped styling of the table; we may or may not want to fix this.

MichaelRoytman:
P.S. I don't know what's going on with the `package-lock.json` file.

src/components/Gradebook/index.jsx
reillz10: The username is guaranteed to be the first item in our functions that format the api data to look good for a table

PR #99, opened 2019-07-02
EDUCATOR-4353 - ui - show override history

Andytr1:


src/components/Gradebook/gradebook.scss
```diff
@@ -98,4 +98,8 @@
 
 .mb-85 {
     margin-bottom: 85px;
+}
+
+.modal-dialog {
```
reillz10: We might want to change the specificity of this. This will change the max-widths of all modals that we can use in our application which isn't desirable. Unfortunately, from the quick look I took, it seems as though the modal component gets placed at the top of the DOM so I'm not exactly sure its possible without editing the Modal component in paragon. @matthugs do you have any thoughts on this?

src/components/Gradebook/index.jsx
```diff
@@ -433,13 +443,40 @@ export default class Gradebook extends React.Component {
                   open={this.state.modalOpen}
                   title="Edit Grades"
                   closeText="Cancel"
-                  body={(
+                  body={( // UI for this component
```
reillz10: Please remove this comment. If you don't that it is clear that the jsx there is for the component, please turn it into a React functional component that way it can be referenced by a name: https://reactjs.org/docs/components-and-props.html

src/components/Gradebook/index.jsx
```diff
@@ -433,13 +443,40 @@ export default class Gradebook extends React.Component {
...
                     <div>
-                      <h3>{this.state.modalModel[0].assignmentName}</h3>
+                      <div>
+                        <div style={{ float: 'left', paddingRight: '49px' }}>Assignment: </div> <div>{this.state.modalAssignmentName}</div>
```
reillz10: Please get rid of the style tag and replace with either some class names or some of bootstraps utility classes. https://getbootstrap.com/docs/4.1/layout/utilities-for-layout/

src/components/Gradebook/index.jsx
```diff
@@ -515,6 +554,7 @@ Gradebook.defaultProps = {
   showBulkManagement: false,
 };
 
+// propTypes are for validation. Like XSD schema
```
reillz10: Please remove this comment.

src/containers/GradebookPage/index.jsx
```diff
@@ -57,16 +64,20 @@ const mapStateToProps = (state, ownProps) => (
   }
 );
 
+// provide these props to be used by component .
+// action creator are on the right
```
reillz10: please remove comments

src/containers/GradebookPage/index.jsx
```diff
@@ -57,16 +64,20 @@ const mapStateToProps = (state, ownProps) => (
...
+  fetchGradeOverrideHistory,
   searchForUser: fetchMatchingUserGrades,
   getPrevNextGrades: fetchPrevNextGrades,
   getCohorts: fetchCohorts,
   getTracks: fetchTracks,
   getAssignmentTypes: fetchAssignmentTypes,
   updateGrades,
   toggleFormat: toggleGradeFormat,
-  filterColumns,
+  filterColumns, // if the property name and fetch name are the same, one liner suffices
```
reillz10: please remove comments

src/data/actions/grades.js
```diff
@@ -97,6 +110,41 @@ const fetchGrades = (
   }
 );
 
+const formatDate = (serverDate) => {
```
reillz10: Move this to a utility file so it can be reused both here and in the template. This is comment #3000 so you can reference in your other file

src/components/Gradebook/index.jsx
```diff
@@ -22,48 +22,56 @@ export default class Gradebook extends React.Component {
...
+     );
+
+     let adjustedGradePossible = '100';
+     if (subsection.attempted) {
+       adjustedGradePossible = ` / ${subsection.score_possible}`;
+     }
+     const options = { year: 'numeric', month: 'long', day: 'numeric' };
+     const timeOptions = { hour: '2-digit', minute: '2-digit' };
+     const dt = new Date();
+     const todaysdate = `${dt.toLocaleDateString('en-US', options)} at ${dt.toLocaleTimeString('en-US', timeOptions)}`;
```
reillz10: Use the formatDate Function that has been moved to a utility file to do this formatting. That comment was #3000

src/data/actions/grades.js
```diff
@@ -97,6 +110,41 @@ const fetchGrades = (
...
+    LmsApiService.fetchGradeOverrideHistory(subsectionId, userId)
+      .then(response => response.data)
+      .then((data) => {
+        dispatch(gotGradeOverrideHistory(
+          formatGradeOverrideForDisplay(data.history), data.override.earned_all_override,
+          data.override.possible_all_override, data.override.earned_graded_override,
+          data.override.possible_graded_override,
+        ));
+      })
+      .catch(() => {
```
reillz10: If the Linter lets you, remove this catch block entirely. If it doesnt at least remove the comments.

src/data/actions/grades.js
```diff
@@ -97,6 +110,41 @@ const fetchGrades = (
...
 );
 
+const formatDate = (serverDate) => {
+  const options = { year: 'numeric', month: 'long', day: 'numeric' };
+  const timeOptions = { hour: '2-digit', minute: '2-digit' };
+  const dt = new Date(serverDate);
+  return `${dt.toLocaleDateString('en-US', options)} at ${dt.toLocaleTimeString('en-US', timeOptions)}`;
+};
+
+// Map from REST response to internal properties
```
reillz10: please remove comments

src/data/actions/grades.js
```diff
@@ -97,6 +110,41 @@ const fetchGrades = (
...
+const formatGradeOverrideForDisplay = historyArray => historyArray.map(item => ({
+  date: formatDate(item.history_date),
+  grader: item.history_user,
+  reason: item.override_reason,
+  adjustedGrade: item.earned_graded_override,
+}));
+
+// Redux action creater wraps web api service and dispatches actions using Redux
```
reillz10: please remove comments

src/data/reducers/grades.js
```diff
@@ -24,6 +31,7 @@ const initialState = {
   showSpinner: true,
 };
 
+// need to dispatch an action with that type.
```
reillz10: remove comments

src/data/reducers/grades.js
```diff
@@ -41,6 +49,24 @@ const grades = (state = initialState, action) => {
         showSpinner: false,
         courseId: action.courseId,
       };
+    case GOT_GRADE_OVERRIDE_HISTORY:
+      return {
+        // create new object,
```
reillz10: Please remove all comments in the reducer

src/components/Gradebook/gradebook.scss
matthugs: Discussed offline using the `parentSelector` prop of paragon's modal component to achieve this (using portals!) https://edx.github.io/paragon/?path=/story/modal--basic-usage

src/data/actions/grades.js
```diff
@@ -97,6 +110,41 @@ const fetchGrades = (
...
+// the inner (return) function is called implicitly by redux -thunk layer
+const fetchGradeOverrideHistory = (subsectionId, userId) => (
+  dispatch =>
+    LmsApiService.fetchGradeOverrideHistory(subsectionId, userId)
+      .then(response => response.data)
+      .then((data) => {
+        dispatch(gotGradeOverrideHistory(
+          formatGradeOverrideForDisplay(data.history), data.override.earned_all_override,
+          data.override.possible_all_override, data.override.earned_graded_override,
+          data.override.possible_graded_override,
```
matthugs: I would pass along the subsectionId & userId we loaded this from as well, so that the reducer can keep track of which combinations have been loaded already, saving us from needing to hit the backend & make the user wait each time they switch which grade they're viewing.

Some pieces of the puzzle to doing that: a second argument is also provided to invoked thunks alongside `dispatch`; `getState` enables you to read from state to determine how you want your thunk to behave. E.g. if we wanted to read from a portion of state to determine whether we need to load data, we can get a reference to state by invoking that function. Simple example here: https://github.com/reduxjs/redux-thunk#motivation

src/data/actions/grades.js
```diff
@@ -39,6 +40,18 @@ const gotGrades = (grades, cohort, track, assignmentType, headings, prev, next,
   courseId,
 });
 
+const gotGradeOverrideHistory = (
+  overrideHistory, currentEarnedAllOverride, currentPossibleAllOverride,
+  currentEarnedGradedOverride, currentPossibleGradedOverride,
```
matthugs: I would recommend using an object parameter to get named arguments, rather than relying on positional arguments. My usual threshold for when to switch from positional to named is about 3 arguments.

More concrete: https://2ality.com/2011/11/keyword-parameters.html

src/data/services/LmsApiService.js
```diff
@@ -87,6 +88,12 @@ class LmsApiService {
       return Promise.reject(result);
     });
   }
+
+  static fetchGradeOverrideHistory(subsectionId, userId) {
+    let historyUrl = `${LmsApiService.baseUrl}/api/grades/v1/subsection/${subsectionId}/?user_id=${userId}`;
+    historyUrl += '&history_record_limit=5';
```
matthugs: Please add these strings together in the initial statement and use `const` rather than `let`

src/components/Gradebook/index.jsx
```diff
@@ -433,13 +443,40 @@ export default class Gradebook extends React.Component {
...
+                        <div style={{ float: 'left', paddingRight: '75px' }}>Student: </div> <div>{this.state.updateUserName}</div>
+                        <div style={{ float: 'left', paddingRight: '25px' }}>Original Grade: </div> <div>{this.state.originalGrade}</div>
+                        <div style={{ float: 'left', paddingRight: '25px' }}>Current Grade: </div> <div>{this.props.gradeOverrideCurrentPossibleGradedOverride}</div>
+                      </div>
                       <Table
-                        columns={[{ label: 'Username', key: 'username' }, { label: 'Current grade', key: 'currentGrade' }, { label: 'Adjusted grade', key: 'adjustedGrade' }]}
-                        data={this.state.modalModel}
+                        columns={[{ label: 'Date', key: 'date' }, { label: 'Grader', key: 'grader' },
+                        { label: 'Reason', key: 'reason' },
+                        { label: 'Adjusted grade', key: 'adjustedGrade' }]}
```
matthugs: Looks like we should move this to a CONSTANT

src/data/actions/grades.js
Andytr1: I understand the idea behind it, but would it be worth doing it in this case, since the function is called with another object's properties that also have descriptive names.

src/components/Gradebook/index.jsx
Andytr1: What would be the benefit of constant? I found an example in the file on line 284 creating a const userInformationHeadingLabel. Is this the same thing?

src/data/actions/grades.js
matthugs: This is absolutely worth doing. It's an extremely easy refactor.
```javascript
const gotGradeOverrideHistory = ({
  overrideHistory, currentEarnedAllOverride, currentPossibleAllOverride,
  currentEarnedGradedOverride, currentPossibleGradedOverride,
  }) => ({
```
and at call sites you just add those same parameter names:
```javascript
        dispatch(gotGradeOverrideHistory({
          overrideHistory: formatGradeOverrideForDisplay(data.history),
          ...
        })
```

src/components/Gradebook/index.jsx
matthugs: To make intention clearer, and to make the code easier to read. It calls out an assumption, and pulls it up to a different point in code where it might be reused/moved to where it could be even more widely reused if necessary.

E.g. we could validate that the props provided to our component doesn't contain data that doesn't fit in these columns. Or we could calculate the column labels & the corresponding propTypes from the same, larger data structure, reducing duplicating this list in multiple places.

src/data/actions/grades.js
Andytr1: Thank you. Will do.

src/data/actions/grades.js
Andytr1: For now, I used the existing pattern of dispatching errorFetching...

src/data/actions/grades.js
Andytr1: Future enhancement, probably

src/components/Gradebook/gradebook.scss
```diff
@@ -42,6 +42,25 @@
         margin-left: 10px;
     }
 }
+.grade-history-header{
+    float: left;
+}
+
+.grade-history-assignment{
+    padding-right: 49px;
```
reillz10: NIT: These class names are super misleading and we would be better suited with more utility classes such as a pr-49 or something like that. That said, this is fine for now and can be refactored later.

src/data/actions/grades.js
reillz10: But why did you do that? It seems as though this changes an errorFetching thing in our store but doesn't result in anything on the UI. Unless I'm missing something here. ...

src/components/Gradebook/index.jsx
```diff
@@ -435,11 +444,46 @@ export default class Gradebook extends React.Component {
...
+                      <Table
+                        columns={GRADE_OVERRIDE_HISTORY_COLUMNS}
+                        data={[...this.props.gradeOverrides, {
+                          date: this.state.todaysDate,
+                          grader: this.state.updateUserName,
+                          reason: (<input
+                            type="text"
+                            name="reasonForChange"
+                            value={this.state.reasonForChange}
+                            onChange={value => this.onChange(value)}
```
matthugs: One teachable moment point (I promise I'll space these out better in the future). If you define the function you're calling like this
```javascript
  onChange = (e) => {
    this.setState({ [e.target.name]: e.target.value });
  }
```
then this can become
```suggestion
                            onChange={this.onChange}
```
and you don't have to create a new function just to call another function with `this` set to the correct thing. This isn't particularly important, but understanding why this works can help you understand the difference between a `function` function and an `=>`/"arrow" function.

src/components/Gradebook/index.jsx
```diff
@@ -13,57 +13,64 @@ import {
...
-          </span>
-        ),
-        assignmentName: `${subsection.subsection_name}`,
-      }],
-      modalOpen: true,
-      updateModuleId: subsection.module_id,
-      updateUserId: userEntry.user_id,
-    });
+  onChange(e) {
+    this.setState({ [e.target.name]: e.target.value });
```
matthugs: Nit: I would prefer a more explicit approach to writing these handlers, i.e. two separate & specifically named methods which handle only their specific use cases and can't be repurposed into something else.

I generally prefer to be very concrete in this sort of context. This method doesn't encapsulate enough to be worth abstracting.

src/components/Gradebook/index.jsx
```diff
@@ -435,11 +444,46 @@ export default class Gradebook extends React.Component {
...
+                                type="text"
+                                name="adjustedGradeValue"
+                                value={this.state.adjustedGradeValue}
+                                onChange={value => this.onChange(value)}
+                              /> {this.state.adjustedGradePossible}
+                            </span>),
+                        }]}
+                      />)}
+
+                      <div>Showing most recent actions(max 5). To see more, please contact
```
matthugs:
```suggestion
                      <div>Showing most recent actions (max 5). To see more, please contact
```

src/data/actions/grades.js
matthugs: I have a formal rule about this sort of code review discussion: if you are pushing off work for the future, show me tickets or I know you're lying ;).

If you could drop something in Jira linking back to this discussion then I'm more than happy to mark this as resolved.

Andytr1: https://openedx.atlassian.net/browse/EDUCATOR-4492

PR #101, opened 2019-07-12
show correct override

Andytr1:


src/components/Gradebook/index.jsx
```diff
@@ -448,38 +447,38 @@ export default class Gradebook extends React.Component {
...
+                              <span>
+                                <input
+                                  type="text"
+                                  name="adjustedGradeValue"
+                                  value={this.state.adjustedGradeValue}
+                                  onChange={value => this.onChange(value)}
+                                /> {this.state.adjustedGradePossible}
+                              </span>),
+                          }]}
+                        />)}
```
Andytr1: This is just whitespace format fixed. The real code change is only on line 557 below

PR #103, opened 2019-07-16
removed username and fake max score

Andytr1:
display correct overrides


src/containers/GradebookPage/index.jsx
```diff
@@ -39,6 +39,10 @@ const mapStateToProps = (state, ownProps) => (
     gradeOverrideCurrentEarnedGradedOverride: state.grades.gradeOverrideCurrentEarnedGradedOverride,
     gradeOverrideCurrentPossibleGradedOverride:
       state.grades.gradeOverrideCurrentPossibleGradedOverride,
+    gradeOriginalEarnedAll: state.grades.gradeOriginalEarnedAll,
```
MichaelRoytman: You don't need to map Redux state to components props that are not used by the component. `gradeOriginalEarnedGraded` is the only piece of Redux state used by the component, so I wouldn't pass the other props.

PR #106, opened 2019-07-22
add selection counts to gradebook

MichaelRoytman:


src/components/Gradebook/index.jsx
```diff
@@ -418,6 +418,13 @@ export default class Gradebook extends React.Component {
                 <small className="form-text text-muted search-help-text">Search by username, email, or student key</small>
               </div>
             </div>
+            <div>
+              <span>Selected </span>
```
matthugs: spans here and on lines 424 & 426 seem redundant

If the whole point is to force spaces, you can do that all in 423 & 425's spans and save some markup noise

src/data/actions/grades.js
```diff
@@ -34,7 +34,10 @@ const startedFetchingGrades = () => ({ type: STARTED_FETCHING_GRADES });
 const finishedFetchingGrades = () => ({ type: FINISHED_FETCHING_GRADES });
 const errorFetchingGrades = () => ({ type: ERROR_FETCHING_GRADES });
 const errorFetchingGradeOverrideHistory = () => ({ type: ERROR_FETCHING_GRADE_OVERRIDE_HISTORY });
-const gotGrades = (grades, cohort, track, assignmentType, headings, prev, next, courseId) => ({
+const gotGrades = (
+  grades, cohort, track, assignmentType, headings, prev,
+  next, courseId, totalUsersCount, filteredUsersCount,
```
matthugs: We probably shouldn't be using positional parameters for all of these; you can switch to what is effectively named parameters by taking a single object parameter, which you destructure using these same names.

PR #107, opened 2019-07-26
EDUCATOR-4532 add button to download interventions

Andytr1:


src/data/services/LmsApiService.js
```diff
@@ -77,6 +77,11 @@ class LmsApiService {
     return downloadUrl;
   }
 
+  static getInterventionExportCsvUrl(courseId) {
+    const downloadUrl = `${LmsApiService.baseUrl}/api/bulk_grades/course/${courseId}/intervention`;
+    return downloadUrl;
```
matthugs: no real reason for this variable name; we can just return this directly.

PR #108, opened 2019-07-29
add filter for specific assignment

matthugs:
JIRA:EDUCATOR-4514

Depends a bit on https://github.com/edx/edx-bulk-grades/pull/13 for all the CSV filters to work appropriately

**Author concerns:** There is a remaining bug with the assignment filter value; because we need additional metadata (the module ID it corresponds to, and the type of assignment) for that filter to work completely consistently, but it's difficult to gauge when we'll have this information in the process of loading the page, there're two outstanding issues with that filter. The first is that the assignment will clear itself inappropriately in a corner case involving reloading the page. The latter is that the assignment filter will not apply to the csv downloaded if the educator doesn't choose a different assignment and switch back, upon loading the page with a saved filter in the URL.


src/data/selectors/filters.js
```diff
@@ -0,0 +1,35 @@
+const getFilters = state => state.filters || {};
+
+const selectableAssignments = (state) => {
+  const selectedAssignmentType = getFilters(state).assignmentType;
+  const needToFilter = selectedAssignmentType && selectedAssignmentType !== 'All';
+  const exampleResultRow = state.grades.results[0] || {};
+  const allAssignments = exampleResultRow.section_breakdown || [];
```
MichaelRoytman: I noticed that the assignment filter disappears from the gradebook when selecting a student group with no users in it but the assignment type filter does not. I understand that the reason that assignment types is still displayed is that the list of assignment types in the course relies on a different endpoint, so we're able to get that data independent of whether or not we have an `exampleResultRow` from the data returned by the grades endpoint, which we use to populate the assignment filter. If there is are no grades data associated with the filters, then I see why assignments would be empty. However, I wonder if might be a better experience to either hide both the assignment type and the assignment drop downs or to persist them and make them disabled in this case. I can imagine a course team member being confused why one drop down disappeared and not the other.

src/data/reducers/filters.js
```diff
@@ -0,0 +1,39 @@
...
+          action.filterType !== 'All' &&
+          (state.assignment || {}).type !== action.filterType)
+          ? 'All' : state.assignment,
+      };
+    case INITIALIZE_FILTERS:
+      return {
+        ...state,
+        ...action.data,
+      };
+    case GOT_GRADES:
```
MichaelRoytman: If I'm understanding correctly, one of the bugs you mentioned involves the fact that if you load the page with query parameters dictating the assignment filter, you don't yet have access to the subsection metadata, like the type and the id, so the state doesn't include that information when the page loads. In that case, could you get that information and add it to the state when the GOT_GRADES action is being handled?

src/data/services/LmsApiService.js
```diff
@@ -70,11 +70,13 @@ class LmsApiService {
   }
 
   static getGradeExportCsvUrl(courseId, options = {}) {
-    const trackQueryParam = options.track ? [`track=${options.track}`] : [];
-    const cohortQueryParam = options.cohort ? [`cohort=${options.cohort}`] : [];
-    const queryParams = [...trackQueryParam, ...cohortQueryParam].join('&');
-    const downloadUrl = `${LmsApiService.baseUrl}/api/bulk_grades/course/${courseId}/?${queryParams}`;
-    return downloadUrl;
+    const queryParams = ['track', 'cohort', 'assignment', 'assignmentType']
+      .filter(opt => Object.prototype.hasOwnProperty.call(options, opt) &&
```
MichaelRoytman: What does the `Object.prototype.hasOwnProperty.call(options, opt)` call tell you? I understand that it checks whether `opt` is a property of the `options` object. But why is that piece of information important? I would think it would be enough to check whether the `opt` key exists in `options`, which it seems like you're doing in the second condition in the expression. Could you explain?

MichaelRoytman: Is there a difference between `Object.prototype.hasOwnProperty.call(options, opt)` and `options.prototype.hasOwnProperty(opt)`? Or are they equivalent?

src/components/Gradebook/index.jsx
```diff
@@ -711,9 +764,17 @@ Gradebook.defaultProps = {
 Gradebook.propTypes = {
   areGradesFrozen: PropTypes.bool,
   assignmentTypes: PropTypes.arrayOf(PropTypes.string),
+  assignmentFilterOptions: PropTypes.arrayOf(PropTypes.shape({
```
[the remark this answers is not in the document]
matthugs: Only if this component depends on those values being set (which it doesn't; it passes them along if they're there but it doesn't depend on them being set).

Ultimately this question is about how stringently we want to validate our prop shape. Generally I tend to be pretty loose about deeply nested prop types, since I find they often change.

src/data/services/LmsApiService.js
matthugs: To your second question, they are not: https://eslint.org/docs/rules/no-prototype-builtins

Generally `options.prototype.hasOwnProperty(opt)` wouldn't do what you may be thinking it'd do. That'd check if the prototype object has the property, not whether `options` has the property.

It is perfectly reasonable to contrast between what I have here and `options.hasOwnProperty(opt)`, and the difference there is that this function could be called with an `options` object of `{ hasOwnProperty: () => true }`, which would end up being called before we got to the prototype implementation. This isn't actually a case where there's any risk of that, but as usual eslint cannot be reasoned with per se.

To your first question, I think I wrote the first condition first and added the other conditions later after I'd plugged this function into its call sites, finding them not covering the cases I needed. I'll remove this condition because it is redundant in context. (There are inputs for which this will behave differently without this condition, but they are highly contrived and involve mucking with prototypes of these objects.)

src/data/actions/filters.js
```diff
@@ -0,0 +1,23 @@
...
+
+const initializeFilters = ({
+  assignment = '',
+  assignmentType = '',
+  track = '',
+  cohort = '',
+}) => ({
+  type: INITIALIZE_FILTERS,
+  data: {
+    assignment: { label: assignment },
```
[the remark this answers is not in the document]
matthugs: You're right that these are poor choices of default values, and thinking a bit harder about them did help me discover & fix a bug. Thanks!

src/data/selectors/filters.js
matthugs: Switched to a disabled state. Their size is different so it doesn't really stop the flicker-fest, but I agree this is a bit better.

src/data/reducers/filters.js
matthugs: Be careful what you wish for.... https://github.com/edx/gradebook/pull/108/files#diff-0b7c15d7052c2ebb52aa2bb06d04566dR28

src/components/Gradebook/index.jsx
```diff
@@ -379,20 +406,32 @@ export default class Gradebook extends React.Component {
...
+                        disabled={this.props.assignmentFilterOptions.length === 0}
+                      />
+                    </div>
+                    <div className="student-filters">
+                      <span className="label">
+                        Assignment:
+                      </span>
+                      <InputSelect
+                        name="assignment-types"
+                        ariaLabel="Assignment Types"
```
MichaelRoytman: I think that the `ariaLabel` prop should be `Assignment` and the `name` prop `assignment`. Was this a copy-paste error, or am I overlooking something?

PR #109, opened 2019-07-31
pull out bulk management tools to main tab and remove redundant funct…

MichaelRoytman:
…ionality across tabs

### DIFF
--- a/src/components/Gradebook/index.jsx
+++ b/src/components/Gradebook/index.jsx
@@ -357,64 +357,41 @@ export default class Gradebook extends React.Component {
             <h1>Gradebook</h1>
             <h3> {this.props.courseId}</h3>
             {this.props.areGradesFrozen &&
-            <div className="alert alert-warning" role="alert" >
-              The grades for this course are now frozen. Editing of grades is no longer allowed.
-            </div>
+              <div className="alert alert-warning" role="alert" >
+                The grades for this course are now frozen. Editing of grades is no longer allowed.
+              </div>
             }
             {(this.props.canUserViewGradebook === false) &&
-            <div className="alert alert-warning" role="alert" >
-              You are not authorized to view the gradebook for this course.
-            </div>
+              <div className="alert alert-warning" role="alert" >
+                You are not authorized to view the gradebook for this course.
+              </div>
             }
             <Tabs labels={this.getActiveTabs()}>
               <div>
+                <h4>Step 1: Filter the Grade Report</h4>
                 <div className="d-flex justify-content-between" >
                   {this.props.showSpinner && <div className="spinner-overlay"><Icon className={['fa', 'fa-spinner', 'fa-spin', 'fa-5x', 'color-black']} /></div>}
-
                   <div>
-                    <div role="radiogroup" aria-labelledby="score-view-group-label">
-                      <span id="score-view-group-label">Score View:</span>
-                      <span>
-                        <label className="mr-2" htmlFor="score-view-percent">
-                          <input
-                            id="score-view-percent"
-                            className="ml-2 mr-1"
-                            type="radio"
-                            name="score-view"
-                            value="percent"
-                            defaultChecked
-                            onClick={() => this.props.toggleFormat('percent')}
-                          />
-                          Percent
-                        </label>
-                      </span>
-                      <span>
-                        <label htmlFor="score-view-absolute">
-                          <input
-                            id="score-view-absolute"
-                            type="radio"
-                            name="score-view"
-                            value="absolute"
-                            className="mr-1"
-                            onClick={() => this.props.toggleFormat('absolute')}
-                          />
-                          Absolute
-                        </label>
-                      </span>
-                    </div>
+                    <InputSelect
+                      label="Score View:"
+                      name="ScoreView"
+                      value="percent"
+                      options={[{ label: 'Percent', value: 'percent' }, { label: 'Absolute', value: 'absolute' }]}
+                      onChange={this.props.toggleFormat}
+                    />
                     {this.props.assignmentTypes.length > 0 &&
-                    <div className="student-filters">
-                      <span className="label">
-                        Assignment Types:
-                      </span>
-                      <InputSelect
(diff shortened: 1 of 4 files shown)


src/components/Gradebook/index.jsx
```diff
@@ -357,14 +357,14 @@ export default class Gradebook extends React.Component {
             <h1>Gradebook</h1>
             <h3> {this.props.courseId}</h3>
             {this.props.areGradesFrozen &&
-            <div className="alert alert-warning" role="alert" >
-              The grades for this course are now frozen. Editing of grades is no longer allowed.
-            </div>
+              <div className="alert alert-warning" role="alert" >
```
MichaelRoytman: I'm not sure why eslint flagged this for me all of a sudden.

src/data/reducers/grades.js
```diff
@@ -35,8 +35,8 @@ const initialState = {
   nextPage: null,
   showSpinner: true,
   bulkManagement: {},
-  totalUsersCount: null,
-  filteredUsersCount: null,
+  totalUsersCount: 0,
```
MichaelRoytman: Without this change, the gradebook displays "Showing of total learners" as it is loading. I thought this was more presentable. I considered hiding the text all together but opted for this for now as it's easier.

src/components/Gradebook/index.jsx
```diff
@@ -471,12 +471,28 @@ export default class Gradebook extends React.Component {
                   onClose={() => this.props.closeBanner()}
                   open={this.props.showSuccess}
                 />
-                <div>
-                  Showing
-                  <span className="font-weight-bold"> {this.props.filteredUsersCount} </span>
-                  of
-                  <span className="font-weight-bold"> {this.props.totalUsersCount} </span>
-                  total learners
+                <div className="d-flex justify-content-between mb-2">
```
MichaelRoytman: I had some trouble getting the alignment where I wanted it, so let me know if this doesn't make any sense. I am happy for some CSS help.

src/components/Gradebook/index.jsx
matthugs: Hmmm... I suspect some version discrepancy nonsense is afoot here. We'll want to watch PRs for any edit warring on this point, and point folks to whatever gets decided on a canonicalized version a la https://edx-internal.slack.com/archives/CD4E4BCM9/p1564432657024900

src/data/reducers/grades.js
matthugs: Hiding the text altogether likely wouldn't be terrifically difficult; it just amounts to adding a strategicly placed `this.props.totalUsersCount ? <jsx> : null` to the Gradebook's `render` method (whether or not we ensure its value is numeric by setting a default value here). I'd prefer that to flickering a lie that there are no learners in the course.

src/components/Gradebook/index.jsx
MichaelRoytman: I spoke to Matt, and this is actually a fine approach.

src/components/Gradebook/index.jsx
```diff
@@ -357,64 +359,41 @@ export default class Gradebook extends React.Component {
...
-                            name="score-view"
-                            value="absolute"
-                            className="mr-1"
-                            onClick={() => this.props.toggleFormat('absolute')}
-                          />
-                          Absolute
-                        </label>
-                      </span>
-                    </div>
+                    <InputSelect
```
MichaelRoytman: I decided to just switch this over to a dropdown to match Christine's wireframes while I was in this area.

src/components/Gradebook/index.jsx
matthugs: I think `align-items-center` is what I was reaching for when we were talking through this in person: that would center the button and the explanatory text. If we want the header and the explanatory text as one thing, I think we'd want to use the same `justify-content` setting as you use here, and use `align-self` on the button to make it more centered relative to the stuff on the other side. And have the heading be inside the first child div of the one on this line.

src/components/Gradebook/index.jsx
```diff
@@ -343,6 +343,8 @@ export default class Gradebook extends React.Component {
     return headings;
   }
 
+  toggleFormat = format => (this.props.toggleFormat(format));
```
matthugs: This may be redundant; you should be able to pass in `this.props.toggleFormat` in place of `this.toggleFormat`.

To be clear there isn't an actual impact to this; only motivation is making the code more concise.

src/containers/GradebookPage/index.jsx
```diff
@@ -68,6 +68,7 @@ const mapStateToProps = (state, ownProps) => (
     uploadSuccess: !!(state.grades.bulkManagement &&
                       state.grades.bulkManagement.uploadSuccess),
     showBulkManagement: stateHasMastersTrack(state),
+    showDownloadButtons: stateHasMastersTrack(state),
```
MichaelRoytman: I couldn't decide whether to add a second, somewhat redundant prop or to  just use the `showBulkManagement` prop for both purposes (perhaps with a more generic name). Thinking it through again, I don't expect that there will be a course where `showBulkManagement` and `showDownloadButtons` would not agree, so maybe a single prop would be better. Any thoughts either way?

matthugs: I prefer it the way you have it in the PR, but can appreciate why it might give you pause. In my mind this way is more explicit, and enables the possibility for a user of the Gradebook component to not have these two distinct flags coincide. It doesn't really matter for this particular use-case, because the Gradebook component is really tightly coupled to its use on this page, but it's still better to orient ourselves towards reusable components whenever possible (since it helps us get the most out of React).

src/data/reducers/grades.js
MichaelRoytman: That makes sense. I have added that.

I originally had mistakenly done `{this.props.totalUsersCount && <jsx>}`, which just displayed a very unhelpful `0`. Now I realize that the left hand sign of the && operator should be true or false, not a truthy or falsy value.

PR #110, opened 2019-08-09
Add assignment grade filter

matthugs:
Depends on: https://github.com/edx/edx-platform/pull/21293 & https://github.com/edx/edx-bulk-grades/pull/15

JIRA:EDUCATOR-4541


src/data/services/LmsApiService.js
```diff
@@ -5,19 +5,36 @@ class LmsApiService {
...
-      gradebookUrl += `cohort_id=${cohort}&`;
+      queryParams.cohort_id = cohort;
     }
     if (track) {
-      gradebookUrl += `enrollment_mode=${track}`;
+      queryParams.enrollment_mode = track;
+    }
+    if (options.assignmentGradeMax || options.assignmentGradeMin) {
+      if (!options.assignment) {
+        throw new Error('Gradebook LMS API requires assignment to be set to filter by min/max assig. grade');
```
jansenk: Why is this?

src/data/actions/grades.js
```diff
@@ -244,6 +264,24 @@ const fetchBulkUpgradeHistory = courseId => (
     }).catch(() => dispatch(bulkHistoryError()))
 );
 
+const updateGradesIfAssigGradeFiltersSet = (
```
iloveagent57: typo: "Assign"

src/data/actions/grades.js
```diff
@@ -244,6 +264,24 @@ const fetchBulkUpgradeHistory = courseId => (
...
+  courseId,
+  cohort,
+  track,
+  assignmentType,
+) => (dispatch, getState) => {
+  const { filters } = getState();
+  const hasAssigGradeFiltersSet = filters.assignmentGradeMax || filters.assignmentGradeMin;
+  if (hasAssigGradeFiltersSet) {
```
iloveagent57: same typo here

src/containers/GradebookPage/index.jsx
```diff
@@ -65,6 +68,16 @@ const mapStateToProps = (state, ownProps) => (
       track: state.filters.track,
       assignment: (state.filters.assignment || {}).id,
       assignmentType: state.filters.assignmentType,
+      assignmentGradeMin: formatMinAssigGrade(
+        state,
+        (state.filters.assignment || {}).id,
```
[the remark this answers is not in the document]
matthugs: Nope, it'll return `undefined`

Try it out in the JS console!

src/data/reducers/filters.js
```diff
@@ -24,11 +24,11 @@ const reducer = (state = initialState, action) => {
       };
     case GOT_GRADES: {
       const { assignment } = state;
-      const { label, type } = assignment || {};
+      const { id, type } = assignment || {};
       if (!type) {
         const relevantAssignment = getAssignmentsFromResultsSubstate(action.grades)
           .map(chooseRelevantAssignmentData)
-          .find(assig => assig.label === label);
+          .find(assig => assig.id === id);
```
iloveagent57: ...and here...now I'm worried that we copypasta'd this everywhere a while back and never caught it.  Maybe it'd be easier to not use this abbreviation?

src/data/services/LmsApiService.js
```diff
@@ -5,19 +5,36 @@ class LmsApiService {
...
+      queryParams.assignment = options.assignment;
+      if (options.assignmentGradeMin) {
+        queryParams.assignment_grade_min = options.assignmentGradeMin;
+      }
+      if (options.assignmentGradeMax) {
+        queryParams.assignment_grade_max = options.assignmentGradeMax;
+      }
     }
+
+    const queryParamString = Object.keys(queryParams)
```
iloveagent57: nice!

src/components/Gradebook/index.jsx
```diff
@@ -131,18 +137,25 @@ export default class Gradebook extends React.Component {
...
-  updateQueryParams = (queryKey, queryValue) => {
+  updateQueryParams = (queryParams) => {
     const parsed = queryString.parse(this.props.location.search);
-    if (queryValue) {
-      parsed[queryKey] = queryValue;
-    } else {
-      delete parsed[queryKey];
-    }
-    return `?${queryString.stringify(parsed)}`;
+    Object.keys(queryParams).forEach((key) => {
```
iloveagent57: nice!

src/data/services/LmsApiService.js
matthugs: Can't filter by assignment grade if you haven't told me which assignment grade to look up. Overall course grade is a separate filter.

The assignment grade filter inputs are also disabled in the UI when the assignment isn't chosen, so we won't get into this situation.

src/data/actions/grades.js
matthugs: Assign is a word in it's own right; I actually find that abbreviation less readable b/c I often misread it as a verb. If this is important to you for consistency's sake then I can do it, but could you confirm

iloveagent57: Hmm, fair point.  I don't want you to change it to something that you have pointed out a valid shortcoming of.  Is it worth using the whole word instead?  We're only saving 5 characters.

iloveagent57: Also, this is probably less of a worry in local variable names than it is for function names, so don't feel like you have to track down every instance of "assig" :)

src/components/Gradebook/index.jsx
```diff
@@ -406,32 +436,70 @@ export default class Gradebook extends React.Component {
...
+                            ariaLabel="Assignment"
+                            value={this.props.selectedAssignment}
+                            options={this.getAssignmentFilterOptions()}
+                            onChange={this.handleAssignmentFilterChange}
+                            disabled={this.props.assignmentFilterOptions.length === 0}
+                          />
+                        </div>
+                        <p>Grade Range (0% - 100%)</p>
+                        <form className="d-flex justify-content-between" onSubmit={this.handleSubmitAssignmentGrade}>
+                          <InputText
```
MichaelRoytman: Olga asked me to include a % sign next to each input - something like this. Could you include that with these inputs too? I imagine she'd want that here as well.

image

src/data/actions/grades.js
matthugs: Sounds great; I'll change all the abbev.s floatin' arou

src/components/Gradebook/index.jsx
```diff
@@ -274,6 +284,26 @@ export default class Gradebook extends React.Component {
     }
   };
 
+  handleSubmitAssignmentGrade = (event) => {
+    event.preventDefault();
+    const formContents = new FormData(event.target);
```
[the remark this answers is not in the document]
matthugs: No you're absolutely right! I only switched to using state after having attempted to write these components as [uncontrolled](https://reactjs.org/docs/uncontrolled-components.html). My hope for doing that was foiled by a paragon limitation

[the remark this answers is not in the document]
matthugs: I'm pretty confident we still support it, certainly for learners. Thanks for the reminder! (FormData isn't altogether verboten for IE11-supporting browsers, but you're right that the functionality I'm using here is! In other situations it can still be used in a cross-browser way to customize form submission behavior without losing the benefit of how the browser will package up the form data for you.) Thanks for pointing this out; strong motivation for making that change to a state data source

src/data/actions/grades.js
```diff
@@ -102,9 +104,27 @@ const fetchGrades = (
   assignmentType,
   options = {},
 ) => (
-  (dispatch) => {
+  (dispatch, getState) => {
     dispatch(startedFetchingGrades());
-    return LmsApiService.fetchGradebookData(courseId, options.searchText || null, cohort, track)
+    const {
```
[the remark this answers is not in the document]
matthugs: I agree directionally with pulling the search string out of state, but I'm not the dictator of gradebook so my preference is by no means law. I'd recommend we socialize that prospect with Neem if you think it's worthwhile to achieve full consistency of approach now.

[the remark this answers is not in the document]
matthugs: It's not immediately clear to me what the motivation was for the original code, but it strikes me as an approach that was fraying a little in its clean applicability to the problem even before we added these additional filters. My perspective is that filters are rapidly becoming something a large featureset in their own right, so 1) it makes sense to split them off into their own slice of state, and 2) it makes sense to decouple the portions of our UI whose behavior is based on the values of those filters from the setting of the filters themselves. That way we can start to introduce more separation in our UI (which is a whole lota' logic in one big heaping component right now) without having to worry about accidentally unthreading this data.

But again I wasn't planning on changing course on all of that immediately.

src/containers/GradebookPage/index.jsx
```diff
@@ -50,6 +51,8 @@ const mapStateToProps = (state, ownProps) => (
     selectedCohort: state.filters.cohort,
     selectedAssignmentType: state.filters.assignmentType,
     selectedAssignment: (state.filters.assignment || {}).label,
+    selectedMinAssigGrade: state.filters.assignmentGradeMin || 0,
```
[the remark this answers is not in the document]
matthugs: Another holdover from when I was trying to make those components uncontrolled.

src/components/Gradebook/index.jsx
matthugs: I can, but if the goal is consistency b/t them it may be better for you to do it, since I can't see your code as of right now.

PR #111, opened 2019-08-13
mroytman/educator 4432 course level grade filter

MichaelRoytman:


src/components/Gradebook/index.jsx
```diff
@@ -48,8 +52,16 @@ export default class Gradebook extends React.Component {
     this.props.initializeFilters(urlQuery);
     this.props.getRoles(this.props.courseId);
     this.overrideReasonInput.focus();
-    const { assignmentGradeMin, assignmentGradeMax } = urlQuery;
-    this.setState({ assignmentGradeMin, assignmentGradeMax });
+
+    // do not override the default values if the query parameters are undefined
+    const courseGradeMin = urlQuery.courseGradeMin || this.state.courseGradeMin;
```
MichaelRoytman: I realize that this has the potential to set the state to the exact same state as before. I know React wold be smart enough to not actually re-render (not sure about the virtual DOM), so I'm not sure if it's an issue or not.

src/components/Gradebook/index.jsx
```diff
@@ -401,6 +424,62 @@ export default class Gradebook extends React.Component {
...
+      this.updateQueryParams({ courseGradeMin, courseGradeMax });
+    }
+  }
+
+  isGradeFilterValueInRange = (value) => {
+    const valueAsInt = parseInt(value, 10);
+    return valueAsInt >= 0 && valueAsInt <= 100;
+  };
+
+  handleAssignmentFilterChange = (assignment) => {
```
MichaelRoytman: I don't know why it thinks I added this?

src/components/Gradebook/index.jsx
matthugs: I would recommend building the object you're going to pass to setState imperatively (I know, ew) a la
```
const newStateFields = {};
['assignmentGradeMin', ... , 'courseGradeMax'].forEach(attr => {
  if (urlQuery[attr]) {
    newStateFields[attr] = urlQuery[attr];
  }
});
this.setState(newStateFields);
```

If you wanted to write this fully functionally, you'd need to use [the functional form of setState](https://stackoverflow.com/a/48209870) since the new version depends on the previous values of state.

src/data/selectors/grades.js
```diff
@@ -95,9 +95,25 @@ const formatMinAssignmentGrade = (state, assignmentId, percentGrade) => {
   return percentGrade;
 };
 
+const formatMaxCourseGrade = (state, percentGrade) => {
```
matthugs: I think we have a functional programming pattern we can extract here: these functions have a common shape which can be its own function. I think we could represent these different iterations of the same idea using function composition as our means of combination, to DRY up this code between how assignment & course grades are handled.

Feel free to bug me if I don't follow up with code on this comment; don't want the description of what I'm after to stay this abstract

matthugs: Also these don't actually need `state`

matthugs:
```
const composeFilters = (...predicates) => (percentGrade, options = {}) =>
  predicates.reduce((accum, predicate) => {if(predicate(percentGrade, options)) {return null;} return accum;});
```
used e.g.
```
formatMaxCourseGrade = composeFilters(percentGradeIsMax);
formatMinAssignmentGrade = composeFilters(
  percentGradeIsMin,
  (percentGrade, { assignmentId }) => !assignmentId,
);
```

PR #113, opened 2019-08-20
Add support for waffling off bulk management

matthugs:
even for masters courses

Depends on https://github.com/edx/edx-platform/pull/21389


src/components/Gradebook/index.jsx
```diff
@@ -683,7 +683,7 @@ export default class Gradebook extends React.Component {
                     options={[{ label: 'Percent', value: 'percent' }, { label: 'Absolute', value: 'absolute' }]}
                     onChange={this.props.toggleFormat}
                   />
-                  {this.props.showDownloadButtons && (
+                  {this.props.showBulkManagement && (
```
matthugs: Collapsing these two props into one, since it's clear they won't change separately in the future. (This wasn't as clear before this change was request.)

src/data/actions/assignmentTypes.js
```diff
@@ -4,12 +4,17 @@ import {
...
   GOT_ARE_GRADES_FROZEN,
 } from '../constants/actionTypes/assignmentTypes';
+import GOT_BULK_MANAGEMENT_CONFIG from '../constants/actionTypes/config';
 import LmsApiService from '../services/LmsApiService';
 
 const startedFetchingAssignmentTypes = () => ({ type: STARTED_FETCHING_ASSIGNMENT_TYPES });
 const errorFetchingAssignmentTypes = () => ({ type: ERROR_FETCHING_ASSIGNMENT_TYPES });
 const gotAssignmentTypes = assignmentTypes => ({ type: GOT_ASSIGNMENT_TYPES, assignmentTypes });
 const gotGradesFrozen = areGradesFrozen => ({ type: GOT_ARE_GRADES_FROZEN, areGradesFrozen });
+const gotBulkManagementConfig = bulkManagementEnabled => ({
```
matthugs: If I was following the pattern super formally, I'd put this in a new `actions` file, like I've done with the action type constants. I've found, however, that grouping actions according by how their dispatched helps to avoid cyclic dependency issues, even though it's a bit less symmetrical/diverges a bit from the pattern of "all the things that correspond to this slice of state go in a file named according to the slice of state."

PR #117, opened 2019-09-06
Push-aside drawer for filters

matthugs:
JIRA: https://openedx.atlassian.net/browse/EDUCATOR-4579

Makes gradebook full-width, moving filters into a drawer which pushes the whole page to the side. The portion of the page still visible to the right of the drawer is left-right scrollable & interactible, enabling us to leverage the previously live-updated filters.

There's a good amount more work that would need to go into pulling this new Drawer component out into paragon:
1. There's some keyboard functionality missing (e.g. ESC-key closing & focus save/restore)
2. Configurability around some hard-coded widths (namely the width of the content that doesn't go in the drawer)

### DIFF
--- a/src/components/Drawer/index.jsx
+++ b/src/components/Drawer/index.jsx
@@ -1,0 +1,86 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { Button } from '@edx/paragon';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faTimes } from '@fortawesome/free-solid-svg-icons';
+
+export default class Drawer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      open: props.initiallyOpen,
+      transitioning: false,
+    };
+  }
+
+  deferToNextRepaint(callback) {
+    window.requestAnimationFrame(() =>
+      window.setTimeout(callback, 0));
+  }
+
+  close = () => {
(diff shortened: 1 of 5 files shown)


src/components/Drawer/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+      this.toggleOpen();
+    }
+  };
+
+  toggleOpen = () => {
+    this.setState({ transitioning: true });
+    // defer the transition to the next repaint so we can be sure that
+    // opening drawer is visible before it transitions
+    // (the start state of the opening animation doesn't work if the element starts hidden)
+    // TODO: this doesn't actually do what I want it to do
```
iloveagent57: Do you want to put a ticket in the backlog to make it do what you really want?  Then we can make sure you have time to do it in the future.

matthugs: thanks for catching this; I did find the issue but forgot to remove the todo

src/components/Drawer/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+            },
+          )}
+          onTransitionEnd={this.handleSlideDone}
+        >
+          <div className="drawer-header">
+            <h2>{this.props.title}</h2>
+            <Button
+              className="p-1"
+              onClick={this.close}
+              aria-label="Close Filters"
```
albemarle: As a note, if Gradebook is internationalized, then a11y text needs to be i18ned as well.

PR #118, opened 2019-09-06
EDUCATOR-4524 - new events

Andytr1:
New events and renamed old ones for google analytics


src/components/Gradebook/index.jsx
```diff
@@ -13,12 +13,15 @@ import {
...
   Tabs,
 } from '@edx/paragon';
+import { trackEvent } from '@redux-beacon/segment';
 import queryString from 'query-string';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faDownload, faSpinner } from '@fortawesome/free-solid-svg-icons';
 import { configuration } from '../../config';
 import PageButtons from '../PageButtons';
 import { formatDateForDisplay } from '../../data/actions/utils';
+import { trackingCategory } from '../../data/store';
```
matthugs: It's not super appropriate for component code to depend directly on the store

Could you move the tracking category into a separate module by itself, to be imported in both places where it's needed?

src/components/Gradebook/index.jsx
```diff
@@ -463,6 +480,13 @@ export default class Gradebook extends React.Component {
         },
       );
       this.updateQueryParams({ courseGradeMin, courseGradeMax });
+      trackEvent(() => ({
```
matthugs: Previously we only had actions which fired off in tandem with a particular redux action was dispatched. I believe we can use the same pattern here, since `updateCourseGradeFilter` does fire a redux action.

src/components/Gradebook/index.jsx
```diff
@@ -280,10 +283,24 @@ export default class Gradebook extends React.Component {
   };
 
   handleClickExportGrades = () => {
+    trackEvent(() => ({
```
matthugs: I'd prefer for these `trackEvent` calls to be moved into the container (i.e. `GradebookPage/index.jsx`) and passed in to the component as new function props, which we can invoke here.

PR #121, opened 2019-09-10
Gradebook styling issues

matthugs:
JIRA: https://openedx.atlassian.net/browse/EDUCATOR-4622
along with an issue I haven't stuck in our bug spreadsheet (described below)

before
Screen Shot 2019-09-09 at 1 57 53 PM
after
Screen Shot 2019-09-10 at 2 02 28 PM
Screen Shot 2019-09-10 at 2 02 04 PM

### DIFF
--- a/src/components/Gradebook/index.jsx
+++ b/src/components/Gradebook/index.jsx
@@ -831,16 +831,14 @@ export default class Gradebook extends React.Component {
                 onChange={this.handleMaxAssigGradeChange}
               />
               <span className="input-percent-label">%</span>
-              <div className="d-flex align-items-center">
-                <Button
-                  type="submit"
-                  className="btn-outline-secondary"
-                  name="assignmentGradeMinMax"
-                  disabled={!this.props.selectedAssignment}
-                >
-                  Apply
-                </Button>
-              </div>
+              <Button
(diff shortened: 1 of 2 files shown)


matthugs:
before for the ticketed bug:
Screen Shot 2019-09-10 at 2 16 05 PM
after:
Screen Shot 2019-09-10 at 2 16 29 PM

src/components/Gradebook/index.jsx
```diff
@@ -831,16 +842,14 @@ export default class Gradebook extends React.Component {
                 onChange={this.handleMaxAssigGradeChange}
               />
               <span className="input-percent-label">%</span>
-              <div className="d-flex align-items-center">
```
iloveagent57: What does getting rid of this `<div>` do?

iloveagent57: Ah, you just added a screenshot that answers this question :)

matthugs: The div was an attempt to get the button into a better-aligned position, but it also complicated efforts to fix this bug. I figure it's also good to make the assignment grade filter and the course grade filter markup consistent with one another, also.

PR #125, opened 2019-10-02
EDUCATOR-4685 add course id to label

Andytr1:


src/data/store.js
```diff
@@ -34,6 +35,7 @@ const eventsMap = {
   })),
   [GRADE_UPDATE_SUCCESS]: trackEvent(action => ({
     name: 'edx.gradebook.grades.grade_override.succeeded',
+    label: action.courseId,
     properties: {
       category: trackingCategory,
       courseId: action.courseId,
```
matthugs: It sounds like we're following through with https://github.com/edx/frontend-app-gradebook/pull/120#discussion_r322946352 adding labels to all of some of these events.

Can we revisit the question of whether we need to duplicate the course id in both the properties & the label?

Andytr1: To give others more context of my conversation with Matt:
I decided to keep the "properties" block as is, because it is being used in segment logs. In order to determine the correct usage/syntax of label, I moved it to line 38. We know that the elements, in existing "properties" block aren't being rendered in google analytics. Therefore, I took a two pronged approach: add "label" inside the properties block as well as outside of it.
Please let me know what changes are desired to move this forward.

PR #133, opened 2019-10-11
add FilterBadges

jansenk:
Added a new component, `FilterBadges`.

For each filter we have, if the filter is active, display a bootstrap Badge under the `Edit Filters` button with the name and value of the filter. Each badge also has a close button, which, when clicked, will clear the filter and any other dependent values.

For min and max, I decided it made the most sense to have them appear as a single 'range' badge than to have individual max and min badges.
Screen Shot 2019-10-11 at 12 57 52 PM
)


src/components/FilterBadges/index.jsx
```diff
@@ -0,0 +1,150 @@
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import initialFilters from '../../data/constants/filters';
+
+/*
+  Active filter values are stored in GradebookPage State, passed into props in selected* variables
```
jansenk: I typed this out for myself as I was working to think through some of it, and I was gonna delete it, but .... maybe it would just be better to leave it here?

src/components/Gradebook/index.jsx
```diff
@@ -571,6 +599,7 @@ export default class Gradebook extends React.Component {
                     <small className="form-text text-muted search-help-text">Search by username, email, or student key</small>
                   </div>
                 </div>
+                {FilterBadges(this.props, this.handleFilterBadgeClose)}
```
matthugs: It's generally considered bad form to pass the entire `props` object to a child. One of the goals when writing React code is to be as explicit as possible about which props are being passed down the component hierarchy, so that components are as independent of one another as possible.

src/components/Gradebook/index.jsx
```diff
@@ -515,6 +517,32 @@ export default class Gradebook extends React.Component {
...
+    const rangeStateFilters = ['assignmentGradeMin', 'assignmentGradeMax', 'courseGradeMin', 'courseGradeMax'];
+    rangeStateFilters.forEach((filterName) => {
+      if (filterNames.includes(filterName)) {
+        stateUpdate[filterName] = initialFilters[filterName];
+      }
+    });
+    this.setState(
+      stateUpdate,
+      () => {
+        this.props.getUserGrades(
```
matthugs: I don't think there's actually a compelling reason these can't happen in parallel, which should help free you up to pull more of this `handleFilterBadgeClose`'s logic into the subcomponent.

src/components/FilterBadges/index.jsx
```diff
@@ -0,0 +1,150 @@
...
+  return (
+    <div>
+      {
+        renderSingleValueFilterBadge(
+          'Assignment Type',
+          'assignmentType',
+          selectedAssignmentType,
+          handleFilterBadgeClose(['assignmentType']),
+        )
+      }
```
matthugs: I think this would be better written in JSX syntax, which can be done by

1. deleting `render` from the front of the names of these functions
2. changing their signature to take a single object argument of props, rather than four positional arguments
3. rewriting as e.g. ```
<SingleValueFilterBadge
  displayName="Assignment Type"
  filterName="'assignmentType"
  filterValue={selectedAssignmentType}
  handleClose={handleFilterBadgeClose(['assignmentType'])}
/>
```

src/components/Gradebook/index.jsx
jansenk: Yeah, you're right. the setState is just for updating the displayed values on the page, not the actual filter values, so I should be good... the filter values in the props should have been updated by resetFilters already, right?

src/components/FilterBadges/index.jsx
matthugs: You should definitely delete this. A subcomponent shouldn't care about how Gradebook and GradebookPage are written, which may well change without this component changing.

src/components/Gradebook/index.jsx
matthugs: Yeah from what I can see as long as we dispatch what's behind `getUserGrades` strictly after we dispatch the action that clears the filters, we should be good.

PR #135, opened 2019-10-18
Upgrade gradebook from paragon 4 to 7

Tj-Tracy:
A minimally-invasive upgrade of paragon from version 4 to 7. I'm happy to say none of the tests broke, and only 1 small UI tweak for spacing's sake. 
image

image

image

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -5,8 +5,6 @@ $fa-font-path: "~font-awesome/fonts";
 @import "~font-awesome/scss/font-awesome";
 
 $input-focus-box-shadow: $input-box-shadow; // hack to get upgrade to paragon 4.0.0 to work
-//@import "~@edx/paragon/src/SearchField/SearchField";
-//@import "~@edx/paragon/src/Collapsible/Collapsible";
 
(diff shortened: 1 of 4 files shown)


src/App.scss
```diff
@@ -1,12 +1,12 @@
...
+@import "~@edx/paragon/scss/edx/theme.scss";
 
 $fa-font-path: "~font-awesome/fonts";
 @import "~font-awesome/scss/font-awesome";
 
 $input-focus-box-shadow: $input-box-shadow; // hack to get upgrade to paragon 4.0.0 to work
-@import "~@edx/paragon/src/SearchField/SearchField";
-@import "~@edx/paragon/src/Collapsible/Collapsible";
+//@import "~@edx/paragon/src/SearchField/SearchField";
+//@import "~@edx/paragon/src/Collapsible/Collapsible";
```
matthugs: red diff best diff

matthugs: (to be clear, I want you to delete these rather than leave them as comments

Tj-Tracy: 🤦‍♂  I never pushed the commit. Thanks for catching that.

PR #136, opened 2019-12-06
Upgrade node build version from 8 to 12

schenedx:
This would upgrade the node build environment of this repo from 8 to 12

@edx/masters-devs Please review as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,7 @@ node_js: 12
 notifications:
   email:
     recipients:
-      - adusenbery@edx.org
-      - rreilly@edx.org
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,9 +1,5 @@
 language: node_js
-node_js:
-  - lts/*
-cache:
-  directories:
-    - "~/.npm"
+node_js: 12
 notifications:
   email:
     recipients:
```
iloveagent57: while you're in here: should we change the recipients to `masters-devs` (or similar)?

PR #138, opened 2020-05-18
Fix aXe output for improperly incrementing header numbers

matthugs:
This is ow aXe reports that error when run from the browser against a non-Master's course gradebook:

Screen Shot 2020-05-18 at 4 32 05 PM

### DIFF
--- a/src/components/GradebookHeader/__snapshots__/test.jsx.snap
+++ b/src/components/GradebookHeader/__snapshots__/test.jsx.snap
@@ -29,9 +29,9 @@ exports[`GradebookHeader component snapshots default values (grades frozen, cann
   <div
     className="subtitle-row d-flex justify-content-between align-items-center"
   >
-    <h3>
+    <h2>
       fakeID
-    </h3>
+    </h2>
   </div>
   <div
     className="alert alert-warning"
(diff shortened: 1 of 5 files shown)


mattcarter:
Are we sure the value change isn't inverted here?  The language used by the tool is ambiguous. Reviewing the DOM, the closest previous sibling heading tag is a `H1`. The described issue "_Heading levels should only increase by one_" signals to me the `H3` heading should be changed from a `H3` into a `H2` heading or the preceding heading changed to a `H2`.  Am I reading between the lines too much?

matthugs:
@mattcarter I confess I did not test against devstack; I made the changes in-browser after a production pageload, reran aXe and noted that the message was resolved. I'd welcome that testing being done on devstack if that's what would satisfy you, but I likely won't have time for a few days to perform it.

mattcarter:
I'm happy to go do the validating. I'll share my results here; likely tomorrow.

muselesscreator:
What was the resolution here?

muselesscreator:
fwiw, from what I read, this seems correct, going from h1->h2->h3.  :shipit:

muselesscreator (approved):

PR #141, opened 2020-08-26
Requirements Update

muselesscreator:
Almost everything in this PR is in service to the changes in `package.json`.

This merge updates our dependencies for the app to what modern MFEs are using elsewhere in the company.

The most pervasive change comes in the form of updating our system to use frontend-platform, which involves different build scripts, linting, and updates to our auth interface, including needing to call/import it differently.

# JIRA: 
https://openedx.atlassian.net/browse/EDUCATOR-5245
https://openedx.atlassian.net/browse/EDUCATOR-5263

FIY: @edx/masters-devs-gta


muselesscreator:
> Wow, this is a lot. I see you put a fair bit of energy not just updating requirements, but making the imports, usage and structures consistent with edX's modern MFEs. I still want to participate in the walkthrough.

TBF: there is a non-trivial amount of this change coming from a lint auto-fix with a more updated set of rules.

PR #142, opened 2020-09-01
display grade override history error from lms

jansenk:
When fetching grade override history, check for new response values `success` and `error_message`, and if `success` is false, display `error_message` rather than the default error message.

Screen Shot 2020-09-01 at 4 04 30 PM

I'm not sure what's up with the spacing on the error message. It looks like the existing error message also looks kinda weird when it doesn't completely span the screen. I'll attach some more screenshots.

@edx/masters-devs-gta 

###  **Important!!!** 
this currently will only work on this branch!! (https://github.com/edx/edx-platform/pull/24849) At least, until the PR is merged

### DIFF
--- a/src/data/actions/grades.test.js
+++ b/src/data/actions/grades.test.js
@@ -5,14 +5,18 @@ import thunk from 'redux-thunk';
 
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { configuration } from '../../config';
-import { fetchGrades } from './grades';
+import { fetchGrades, fetchGradeOverrideHistory } from './grades';
 import {
   STARTED_FETCHING_GRADES,
   FINISHED_FETCHING_GRADES,
   ERROR_FETCHING_GRADES,
   GOT_GRADES,
+  GOT_GRADE_OVERRIDE_HISTORY,
+  ERROR_FETCHING_GRADE_OVERRIDE_HISTORY
 } from '../constants/actionTypes/grades';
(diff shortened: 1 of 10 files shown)


jansenk:
Screen Shot 2020-09-01 at 4 10 58 PM

jansenk:
It certainly squishes a bit weirdly. Probably a new ticket! 
Screen Shot 2020-09-01 at 5 02 52 PM

src/data/actions/grades.test.js
```diff
@@ -176,4 +180,92 @@ describe('actions', () => {
...
+          errorMessage: responseData.error_message,
+        }];
+        const store = mockStore();
+
+        return store.dispatch(fetchGradeOverrideHistory(subsectionId, userId)).then(() => {
+          expect(store.getActions()).toEqual(expectedActions);
+        });
+      });
+
+      it('on 500 error', () => {
```
muselesscreator, line 256: nit: I would use `test('on 500 error')` or describe more in the title what the expected behavior is.

src/data/reducers/grades.test.js
```diff
@@ -161,4 +162,17 @@ describe('grades reducer', () => {
       type: ERROR_FETCHING_GRADES,
     })).toEqual(expected);
   });
+
+  describe('updates fetch grade history failure state', () => {
```
nsprenkle, line 166: I would keep this as an `it`:
```suggestion
it('updates fetch grade history failure state', () => {
```

jansenk: good catch, this is left over from something else i tried

nsprenkle (approved):
Have verified correct error banner behavior.

PR #144, opened 2020-10-19
Locked header and component split-out

muselesscreator:
This PR does 2 things.

# Css Update for locked header row
Css update for table component to allow locked header bar.  Is not a 100% solution, as there are some eccentricities to tables that are a bit wonky for our 100% desired solution.

**NOTE**: PR includes a hack to add a bunch of users into the gradebook table for visibillity of the CSS change.  This is expected to be removed before merge

# Phase 1 of component split-out from the Gradebook component.
This PR pulls a number of sub-components out of the Gradebook mega-component.  The final location and implementation of this pull is not solid yet, as this is a phase one of a more concerted future effort, but this split-out should allow these sub-components to be appropriately tested and managed.

JIRA: [JIRA-5347](https://openedx.atlassian.net/browse/EDUCATOR-53470)
FIY: @edx/masters-devs-gta

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
 NODE_ENV='production',
+NODE_PATH=./src
 BASE_URL=null,
 LMS_BASE_URL=null,
 LOGIN_URL=null,
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ dist/
 
 ### Emacs ###
 *~
+*.swo
+*.swp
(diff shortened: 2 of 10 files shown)


Andytr1:
Just by reading PR comments - are the "two things" that this PR does too closely related and can't be separated?
I would think it makes sense having individual PRs for each discrete change:

- Fixing the ticket
-Component split

jansenk (changes requested):
Manual review picked up a couple of concerns:

1) When I search by username, I see multiple rows with the same name even though there's only one 'edx' user in the course.
many edxes

 Also, I verified in the network panel that the LMS is only returning the one result. 
network panel

Also, when I attempt to type in the override modal, it swallows my inputs and [Object object] pops up in the text box

object object


Perhaps you could do andrews first suggestion and split the two PRs?  That way we can get the scrolling in and fix the bugs in a separate PR

jansenk (approved):
I can't find anything wrong with this

Andytr1 (approved):
lgtm

PR #146, opened 2020-10-21
lock header row and column

muselesscreator:
# Css Update for locked header row
Css update for table component to allow locked header bar.  Is not a 100% solution, as there are some eccentricities to tables that are a bit wonky for our 100% desired solution.

Note: this is just the css from PR 144, which had some somewhat unrelated JS changes associated.

JIRA: [JIRA-5347](https://openedx.atlassian.net/browse/EDUCATOR-53470)
FIY: @edx/masters-devs-gta

### DIFF
--- a/src/components/Gradebook/gradebook.scss
+++ b/src/components/Gradebook/gradebook.scss
@@ -57,9 +57,17 @@
 .grade-history-current-grade{
     padding-right: 25px;
 }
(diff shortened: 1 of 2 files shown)


jansenk (approved):

mattcarter (approved):
looked good to me, functioned as expected

PR #149, opened 2020-11-18
Revert "fix: Update frontend-build and frontend-platform"

coryleeio:
Reverts edx/frontend-app-gradebook#145

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -145,36 +145,14 @@
       }
     },
     "@babel/helper-builder-react-jsx-experimental": {
-      "version": "7.12.4",
-      "resolved": "https://registry.npmjs.org/@babel/helper-builder-react-jsx-experimental/-/helper-builder-react-jsx-experimental-7.12.4.tgz",
-      "integrity": "sha512-AjEa0jrQqNk7eDQOo0pTfUOwQBMF+xVqrausQwT9/rTKy0g04ggFNaJpaE09IQMn9yExluigWMJcj0WC7bq+Og==",
+      "version": "7.10.5",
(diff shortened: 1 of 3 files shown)


morenol:
Hi @jansenk @coryleeio, could you describe the error that you had? maybe we should apply this for the Koa release

nedbat:
@davidjoy I think you know something about this?  Where do we stand, and how do we keep Koa in sync with the master branches so that both work?

davidjoy:
Thanks for following up @morenol and @nedbat - I'm not actually sure what the specific error was, but I'm assuming it's related to the usage of PUBLIC_PATH for the router, as it was a runtime issue.  

My guess is that a bad (empty?) value for PUBLIC_PATH made it in, and caused the router to not match any paths, so the MFE displayed no content.

jansenk:
@morenol @davidjoy @nedbat 
Here are some screenshots from when we were triaging.PUBLIC_PATH was undefined (or null?) and we were trying to call concat on it.

8DEA3AA1-6C5E-4D9D-A4E0-B61F9A946FAD
2A5DB45D-3F50-476C-B13F-BF36D1626880

morenol:
@davidjoy @jansenk then a reasonable fix would be to get that configuration from getConfig of frontend-platform?

import { getConfig } from '@edx/frontend-platform';

{getConfig().PUBLIC_PATH.concat(':courseId')}

right?

davidjoy:
The problem appears to be that PUBLIC_PATH is undefined or null, as opposed to a string. Does the version in getConfig turn it into an empty string if it’s not defined?  If so, that’d work. 

Otherwise, I’d suggest just adding some defensive checking here where we want to use it.

morenol:
Yes, it has a default there:

https://github.com/edx/frontend-platform/blob/b909372a86f4abfed5f61578f1a9e1eea55b3826/src/config.js#L36

PR #154, opened 2020-12-02
logo update

muselesscreator:
Updates the logo URL to the amazon S3 logo for the upcoming branding update.

The only uses in our project seem to be the favicon and the logo in the header.


.env.development
```diff
@@ -4,6 +4,10 @@ BASE_URL='localhost:1994'
 LMS_BASE_URL='http://localhost:18000'
 LOGIN_URL='http://localhost:18000/login'
 LOGOUT_URL='http://localhost:18000/login'
+LOGO_URL=https://edx-cdn.org/v3/default/logo.svg
```
jansenk, line 7: do these have to be updated in other envs too?

muselesscreator: Not according to the Flip Instructions doc.  Just .env.development (and .env.test if applicable, but we don't have that one)
https://openedx.atlassian.net/wiki/spaces/AC/pages/2047345474/Logo+Flip+Plan+and+Instructions#MFE-Logo-Instructions

mattcarter (approved):
looks good, matches spec.  Approved

.env.development
jansenk: neat-o. I'm definitely gonna check out that doc

jansenk (approved):
Since the switchover hasn't happened yet, I don't think there's anything I can test locally, but this all looks right. Looks like we'll be able to test on stage on the 10th

muselesscreator:
FWIW the test is just that the logo appears appropriately in the header and favicon, as if it were not working the image would be broken there

.env.development
nsprenkle, line 7: Nit: follow the quoting scheme that the project already uses. It's jarring to see warring styles.

muselesscreator: counter-nit: this is explicitly copied from the logo-swap doc, which has the https values unquoted.

PR #156, opened 2020-12-10
fix column header width to static value

muselesscreator:
Use width, instead of min-width for columns to ensure same-size-iness
FIY: @edx/masters-devs-gta


mattcarter (approved):
Approved.

jnlapierre (approved):
LGTM

PR #158, opened 2020-12-16
Bulk mgmt tab fix

muselesscreator:
pull Tab wrapper out of BulkManagement component to make react less unhappy with prop-passing


nsprenkle (approved):
Verified locally :shipit:

PR #161, opened 2021-01-05
update axios and frontend-platform

muselesscreator:
update axios and frontend-platform which depends on it to resolve a security issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "@edx/brand": "npm:@edx/brand-edx.org@^1.3.2",
     "@edx/frontend-component-footer": "10.1.1",
-    "@edx/frontend-platform": "^1.8.0",
+    "@edx/frontend-platform": "1.8.1",
     "@edx/paragon": "12.4.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@edx/brand": "npm:@edx/brand-edx.org@^1.3.2",
     "@edx/frontend-component-footer": "10.1.1",
-    "@edx/frontend-platform": "^1.8.0",
+    "@edx/frontend-platform": "1.8.1",
```
nsprenkle, line 31: Is there a reason to pin this at `1.8.1` instead of `^1.8.1`?

nsprenkle: LGTM pending an answer on this line.

nsprenkle (approved):

package.json
mattcarter: Let's keep that convo moving, but elsewhere. I don't think pinning/not pinning is a block to removing the vulnerable version.

mattcarter (approved):
LGTM.

PR #165, opened 2021-01-25
Allow special access to bulk management tools

kdmccormick:
Access is configured on a per-course-run basis
via the new setting:
`BULK_MANAGEMENT_SPECIAL_ACCESS_COURSE_IDS`

See related edx-internal PR: https://github.com/edx/edx-internal/pull/4278

https://openedx.atlassian.net/browse/TNL-7901


stvstnfrd (approved):
:+1:

PR #173, opened 2021-04-21
Fix filters2

muselesscreator:
**TL;DR -**
Filters were not updating query params

JIRA: [JIRA-5696](https://openedx.atlassian.net/browse/JIRA-5696)

**What changed?**
Filters all update query params

**Developer Checklist**
- [x] Test suites passing
- [ ] Bumped version number [package.json](../package.json)

**Testing Instructions**
update filters, verify they update the URI.
refresh and make sure they all stick.

**Reviewer Checklist**

FYI: @edx/masters-devs-gta

### DIFF
--- a/src/data/actions/filters.js
+++ b/src/data/actions/filters.js
@@ -31,7 +31,7 @@ const initializeFilters = ({
     assignmentGradeMax,
     courseGradeMin,
     courseGradeMax,
-    includeCourseRoleMembers,
+    includeCourseRoleMembers: Boolean(includeCourseRoleMembers),
   },
 });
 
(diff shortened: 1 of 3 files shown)


src/data/actions/filters.js
```diff
@@ -31,7 +31,7 @@ const initializeFilters = ({
     assignmentGradeMax,
     courseGradeMin,
     courseGradeMax,
-    includeCourseRoleMembers,
```
jansenk, line 34: What does this do?

nsprenkle (approved):
Beautiful. Have verified locally :shipit: 

_Note:_ Right now the search box doesn't save to the query params (and maybe it shouldn't anyways), but I wrote a separate ticket this morning to note that.

jansenk (commented):
Tested locally, looks good but I still am seeing the bug WRT/ cohorts

src/data/actions/filters.js
muselesscreator: clears a complaint that the value is being passed from urlQuery as string rather than boolean.

PR #174, opened 2021-04-22
Split SearchControls into separate component

nsprenkle:
**TL;DR -** As part of the push to break up the Gradebook monolith, split the "search controls" into a separate component.

JIRA: [EDUCATOR-5663](https://openedx.atlassian.net/browse/EDUCATOR-5663)

**What changed?**

- Search controls (edit filters button and search box, highlighted in screenshot) moved to a new component
- Added test config and unit/snapshot tests for new components 😄 
- Updated package-lock format to new version

Screen Shot 2021-04-22 at 11 47 29

**Developer Checklist**
- [x] Test suites passing

**Testing Instructions**

See Gradebook test plan. Specifically, verify that the following workflows still work as intended:
- [x] Grades table results can be filtered from the "Filter" panel.
- [x] Users can be searched/filtered using the Search box.

**Reviewer Checklist**

FYI: @edx/masters-devs-gta


src/components/Gradebook/SearchControls.jsx
```diff
@@ -0,0 +1,112 @@
...
+            <div className="spinner-overlay">
+              <Icon className="fa fa-spinner fa-spin fa-5x color-black" />
+            </div>
+          )}
+          <Button
+            id="edit-filters-btn"
+            className="btn-primary align-self-start"
+            onClick={this.props.toggleFilterDrawer}
+          >
+            <FontAwesomeIcon icon={faFilter} />Edit Filters
```
muselesscreator, line 60: Why 2 different icon components?

src/components/Gradebook/SearchControls.test.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import { SearchControls } from './SearchControls';
+
```
muselesscreator, line 5: I would mock the Icons, as they are cluttering up the snapshot.
```
jest.mock('@edx/paragon', () => ({
  Icon: 'Icon',
  Button: 'Button',
  SearchField: 'SearchField',
}));
```

src/components/Gradebook/SearchControls.test.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import { SearchControls } from './SearchControls';
```
jansenk, line 4: In ben's PR he also tested mapDispatchToProps and mapStateToProps, so I think we should as well

src/components/Gradebook/SearchControls.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    props = {
+      courseId: 'course-v1:edX+DEV101+T1',
+      filterValue: 'alice',
+      setGradebookState: jest.fn().mockName('setGradebookState'),
+      showSpinner: false,
+      toggleFilterDrawer: jest.fn().mockName('toggleFilterDrawer'),
+      // From Redux
```
muselesscreator, line 17: fwiw, for tests: I find it more helpful to group the props by type (fn vs static) vs from origin, as that makes it easier to see which things need resets, names, etc.
Also: only functions that appear directly in the JSX need names :-)

src/components/Gradebook/SearchControls.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+        const wrapper = searchControls();
+        wrapper.instance().onChange('bob');
+        expect(props.setGradebookState).toHaveBeenCalledWith({ filterValue: 'bob' });
+      });
+    });
+
+    describe('onClear', () => {
+      it('re-runs search with existing filters', () => {
+        const wrapper = searchControls();
+        wrapper.instance().onClear('bob');
```
jansenk, line 65: onclear doesn't take params

src/components/Gradebook/SearchControls.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+        expect(wrapper.find('.spinner-overlay').exists()).toEqual(false);
+      });
+    });
+
+    describe('onSubmit', () => {
+      it('calls props.searchForUser with correct data', () => {
+        const wrapper = searchControls();
+        wrapper.instance().onSubmit('bob');
+
+        expect(props.searchForUser).toHaveBeenCalledWith('course-v1:edX+DEV101+T1', 'bob', 'spring term', 'masters', 'homework');
```
muselesscreator, line 50: toHaveBeenCalledWith args should be split out to their own lines.

muselesscreator: *same comment for other places with multi-arg call checks below.  don't want to spam

src/components/Gradebook/SearchControls.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+        wrapper.instance().onChange('bob');
+        expect(props.setGradebookState).toHaveBeenCalledWith({ filterValue: 'bob' });
+      });
+    });
+
+    describe('onClear', () => {
+      it('re-runs search with existing filters', () => {
+        const wrapper = searchControls();
+        wrapper.instance().onClear('bob');
+        expect(props.getUserGrades).toHaveBeenCalledWith('course-v1:edX+DEV101+T1', 'spring term', 'masters', 'homework');
```
muselesscreator, line 66: split args, plz

src/components/Gradebook/SearchControls.test.jsx
jansenk, line 50: rather than the strings, should we just have `props.courseID` etc?

jansenk (commented):
Haven't tested manually, but my only actual suggestion is not including the spinner. It doesn't belong in SearchControls.

src/components/Gradebook/index.jsx
```diff
@@ -289,32 +289,13 @@ export default class Gradebook extends React.Component {
...
-                      )}
-                      value={this.state.filterValue}
-                    />
-                    <small className="form-text text-muted search-help-text">Search by username, email, or student key</small>
-                  </div>
-                </div>
+                <SearchControls
+                  courseId={this.props.courseId}
+                  filterValue={this.state.filterValue}
+                  setGradebookState={this.safeSetState}
```
muselesscreator, line 295: For a single setter, I would rather use createStateFieldSetter('filterValue'), rather than add to the big bundle.

src/components/Gradebook/SearchControls.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+      getUserGrades: jest.fn().mockName('getUserGrades'),
+      searchForUser: jest.fn().mockName('searchForUser'),
+      selectedAssignmentType: 'homework',
+      selectedCohort: 'spring term',
+      selectedTrack: 'masters',
+    };
+  });
+
+  const searchControls = (overriddenProps) => {
+    props = Object.assign(props, overriddenProps);
```
muselesscreator, line 27: es6 version: `props = { ...props, ...overriddenProps };`

src/components/Gradebook/SearchControls.jsx
```diff
@@ -0,0 +1,112 @@
...
+  fetchMatchingUserGrades,
+} from '../../data/actions/grades';
+
+/**
+ * Controls for filtering the GradebookTable. Contains the "Edit Filters" button for opening the filter drawer
+ * as well as the search box for searching by username/email.
+ */
+export class SearchControls extends React.Component {
+  /** Submitting searches for user matching the username/email in `value` */
+  onSubmit = value => {
```
muselesscreator, line 20: Because these are arrow functions instead of bound class-methods, we can't make them show up in the snapshot with mockNames. :-(

src/components/Gradebook/SearchControls.test.jsx
nsprenkle: So I basically limit this styling to what ESLint complains about (since I have it integrated with my IDE). Valid nit but still a nit.

nsprenkle: Would prefer we have an enforced, auto-runnable style system so we don't have to have to spend time with this but that's a separate issue for the frontend people to figure out 🙃

jest.config.js
```diff
@@ -0,0 +1,11 @@
+const { createConfig } = require('@edx/frontend-build');
```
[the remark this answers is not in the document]
nsprenkle: I stole this from https://github.com/muselesscreator/frontend-app-gradebook/commit/5a2d5d499fe9b8d3ab48caac2e013ea266da187e

src/components/Gradebook/SearchControls.jsx
```diff
@@ -0,0 +1,112 @@
...
+    this.props.searchForUser(
+      this.props.courseId,
+      value,
+      this.props.selectedCohort,
+      this.props.selectedTrack,
+      this.props.selectedAssignmentType,
+    );
+  }
+
+  /** Changing the search value stores the key in Gradebook. Currently unused */
```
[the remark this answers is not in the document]
nsprenkle: Nothing else uses this value, as far as I can tell. It's resident in state, but we only read it from the search box directly when we do a search.

Admittedly, this is probably what causes the bug where modifying a filter when a search is active causes the search value to be ignored but it is current behavior.

src/components/Gradebook/SearchControls.jsx
```diff
@@ -0,0 +1,112 @@
...
+    );
+  }
+
+  render() {
+    return (
+      <>
+        <h4>Step 1: Filter the Grade Report</h4>
+        <div className="d-flex justify-content-between">
+          {this.props.showSpinner && (
+            <div className="spinner-overlay">
```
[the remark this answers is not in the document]
nsprenkle: Yeah, that's fair.

src/components/Gradebook/__snapshots__/SearchControls.test.jsx.snap
```diff
@@ -0,0 +1,37 @@
...
+    >
+      <Icon
+        className="fa fa-filter"
+      />
+       Edit Filters
+    </Button>
+    <div>
+      <SearchField
+        inputLabel="Search for a learner"
+        onChange={[Function]}
```
muselesscreator, line 24: these are showing as anonymous because they are arrow functions and not class methods which are mock-able

nsprenkle: When I change them to class methods it causes the page to break. Not sure why though...

src/components/Gradebook/SearchControls.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+    describe('mapStateToProps', () => {
+      const state = {
+        filters: {
+          assignmentType: 'labs',
+          track: 'honor',
+          cohort: 'fall term',
+        },
+      };
+
+      it('is maps assignment type filter correctly', () => {
```
jansenk, line 85: nit: "is maps"?

jansenk (approved):
LGTM! But ofc i defer to ben

jnlapierre (approved):
Wondering if Ben's request should be a separate ticket?  Really enjoyed walking through the tests as an example of component testing.

PR #175, opened 2021-04-23
break out status alerts into their own component

jansenk:
**TL;DR -** break out status alerts into their own component

JIRA: [EDUCATOR-5664](https://openedx.atlassian.net/browse/EDUCATOR-5664)

**Developer Checklist**
- [x] Test suites passing

**Testing Instructions**

Regression testing:
- All of the following should result in an error status alert appearing:
  - enter a negative value for min course grade and click apply
  - enter a value over 100 for min course grade and click apply
  - enter a negative value for max course grade and click apply
  - enter a value over 100 for max course grade and click apply
  - enter a value over 100 for both max course grade and min course grade and click apply
  - enter a negative value for both max course grade and min course grade and click apply
  - enter a value over 100 for max course grade and a negative value for min course grade and click apply
  - enter a negative value for max course grade and a value over 100 for min course grade and click apply
- Override a grade. After the spinner disappears, the success banner should appear. Performing other actions should not result in the banner disappearing. Clicking the X should remove the banner.

**Reviewer Checklist**

FYI: @edx/masters-devs-gta

### DIFF
--- a/src/components/Gradebook/StatusAlerts.jsx
+++ b/src/components/Gradebook/StatusAlerts.jsx
@@ -1,0 +1,71 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+import { StatusAlert } from '@edx/paragon';
+
+import { closeBanner } from '../../data/actions/grades';
+
+export const maxCourseGradeInvalidMessage = 'Maximum course grade value must be between 0 and 100. ';
+export const minCourseGradeInvalidMessage = 'Minimum course grade value must be between 0 and 100. ';
+
+export class StatusAlerts extends React.Component {
+  get isCourseGradeFilterAlertOpen() {
+    const r = !this.props.isMinCourseGradeFilterValid
+    || !this.props.isMaxCourseGradeFilterValid;
+    return r;
+  }
+
+  get courseGradeFilterAlertDialogText() {
+    let dialogText = '';
+    if (!this.props.isMinCourseGradeFilterValid) {
+      dialogText += minCourseGradeInvalidMessage;
+    }
+    if (!this.props.isMaxCourseGradeFilterValid) {
+      dialogText += maxCourseGradeInvalidMessage;
+    }
+    return dialogText;
+  }
+
+  render() {
+    return (
+      <>
+        <StatusAlert
+          alertType="success"
+          dialog="The grade has been successfully edited. You may see a slight delay before updates appear in the Gradebook."
+          onClose={this.props.handleCloseSuccessBanner}
+          open={this.props.showSuccessBanner}
+        />
+        <StatusAlert
+          alertType="danger"
+          dialog={this.courseGradeFilterAlertDialogText}
+          dismissible={false}
+          open={this.isCourseGradeFilterAlertOpen}
+        />
+      </>
+    );
+  }
+}
+
+StatusAlerts.defaultProps = {
+  isMinCourseGradeFilterValid: true,
+  isMaxCourseGradeFilterValid: true,
+};
+
+StatusAlerts.propTypes = {
+  isMinCourseGradeFilterValid: PropTypes.bool,
+  isMaxCourseGradeFilterValid: PropTypes.bool,
+  // redux
+  handleCloseSuccessBanner: PropTypes.func.isRequired,
+  showSuccessBanner: PropTypes.bool.isRequired,
+};
+
+export const mapStateToProps = (state) => ({
+  showSuccessBanner: state.grades.showSuccess,
+});
+
+export const mapDispatchToProps = {
+  handleCloseSuccessBanner: closeBanner,
+};
+
+export default connect(mapStateToProps, mapDispatchToProps)(StatusAlerts);
--- a/src/components/Gradebook/StatusAlerts.test.jsx
+++ b/src/components/Gradebook/StatusAlerts.test.jsx
@@ -1,0 +1,99 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import {
+  StatusAlerts,
+  mapDispatchToProps,
+  mapStateToProps,
+  maxCourseGradeInvalidMessage,
+  minCourseGradeInvalidMessage,
+} from './StatusAlerts';
+import { closeBanner } from '../../data/actions/grades';
+
+jest.mock('@edx/paragon', () => ({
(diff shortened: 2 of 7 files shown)


src/components/Gradebook/StatusAlerts.test.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { StatusAlerts, mapDispatchToProps, mapStateToProps } from './StatusAlerts';
+import { closeBanner } from '../../data/actions/grades';
+
+describe('StatusAlerts', () => {
+  const renderStatusAlerts = (params) => {
```
nsprenkle, line 7: This is a cool way to assert snapshots in your render test 🆒

nsprenkle (commented):
From a read-through, could probably use a test for `getCourseGradeFilterAlertDialog` but everything else looks good. Will pull and regression test locally.

src/components/Gradebook/StatusAlerts.jsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+import { StatusAlert } from '@edx/paragon';
+
+import { closeBanner } from '../../data/actions/grades';
+
+export class StatusAlerts extends React.Component {
+  getCourseGradeFilterAlertDialog() {
```
muselesscreator, line 10: I don't like using Dialog here as that is an overloaded term that generally means something akin to a modal in this space.
Also, this could be a getter, as that's how you're using it :-)
option: `get courseGradeFilterAlertText` or `get courseGradeFilterAlertDialogText`

src/components/Gradebook/StatusAlerts.test.jsx
```diff
@@ -0,0 +1,66 @@
...
+        isMinCourseGradeFilterValid: true,
+        isMaxCourseGradeFilterValid: true,
+      });
+    });
+
+    it.each([
+      [false, true],
+      [true, false],
+      [false, false],
+    ])('rendered when course grade filters are invalid.', (isMinCourseGradeFilterValid, isMaxCourseGradeFilterValid) => {
```
muselesscreator, line 29: I don't think I've ever seen this syntax... is cool!

src/components/Gradebook/StatusAlerts.test.jsx
```diff
@@ -0,0 +1,66 @@
...
+import { closeBanner } from '../../data/actions/grades';
+
+describe('StatusAlerts', () => {
+  const renderStatusAlerts = (params) => {
+    const alertParams = {
+      handleCloseSuccessBanner: jest.fn().mockName('handleCloseSuccessBanner'),
+      showSuccessBanner: false,
+      ...params,
+    };
+    const alerts = renderer.create(<StatusAlerts {...alertParams} />);
```
muselesscreator, line 13: FWIW, I much much prefer enzyme.shallow over react-test-renderer.  Does not try to interpret the children out, which makes for cleaner snapshots.  (also supports mock-naming better, but that isn't an issue for this component)

src/components/Gradebook/StatusAlerts.test.jsx
```diff
@@ -0,0 +1,66 @@
...
+    it('not rendered otherwise', () => {
+      renderStatusAlerts({
+        showSuccessBanner: false,
+      });
+    });
+  });
+
+  describe('mapStateToProps', () => {
+    it('showSuccessBanner', () => {
+      expect(mapStateToProps({ grades: { showSuccess: true } }).showSuccessBanner).toBeTruthy();
```
muselesscreator, line 53: This is a place where I would just set a weird value and track across.
Alternatively, I have set up in the past a set of libraries for checking selectors that are a bit more brute-force like this, but wrapped in libraries.

muselesscreator (changes requested):
Looks good, but I would like us to rely on one rendering engine for tests wherever possible (enzyme)

nsprenkle:
Local testing looks good ✅  but will defer to @muselesscreator for stylistic/JS best-practices.

jansenk:
> From a read-through, could probably use a test for `getCourseGradeFilterAlertDialog` but everything else looks good. Will pull and regression test locally.

I could unit test it, but all branches are already covered by the snapshots, so I assumed that was sufficient

src/components/Gradebook/StatusAlerts.jsx
```diff
@@ -0,0 +1,65 @@
...
+          dialog="The grade has been successfully edited. You may see a slight delay before updates appear in the Gradebook."
+          onClose={this.props.handleCloseSuccessBanner}
+          open={this.props.showSuccessBanner}
+        />
+        <StatusAlert
+          alertType="danger"
+          dialog={this.courseGradeFilterAlertDialogText}
+          dismissible={false}
+          open={
+            !this.props.isMinCourseGradeFilterValid
```
muselesscreator, line 35: If you put this check in a getter, you can do a single snapshot, and then a logic test on the method (rather than 3 snapshots)

muselesscreator (approved):

src/components/Gradebook/StatusAlerts.jsx
```diff
@@ -0,0 +1,65 @@
...
+    }
+    if (!this.props.isMaxCourseGradeFilterValid) {
+      dialogText += 'Maximum course grade value must be between 0 and 100. ';
+    }
+    return dialogText;
+  }
+
+  render() {
+    return (
+      <div>
```
nsprenkle, line 23: Nit: this can probably be a react fragment `<>` instead of a div to avoid any changes to the output HTML.

nsprenkle (approved):
Looks good. Have run smoke test locally.

src/components/Gradebook/StatusAlerts.test.jsx
```diff
@@ -0,0 +1,99 @@
...
+        ).toEqual(
+          expect.stringContaining(minCourseGradeInvalidMessage),
+        );
+      }
+    });
+  });
+
+  describe('mapStateToProps', () => {
+    it('showSuccessBanner', () => {
+      const arbitraryValue = 'AppleBananaCucumber';
```
muselesscreator, line 81: <3

muselesscreator (approved):
Changes look great!  Nice looking snapshot :-)

PR #200, opened 2021-06-30
Simplify Bulk Management Config

nsprenkle:
**TL;DR -** outsource bulk management enabling logic to edx-platform.

JIRA: 
* Drafted as part of [EDUCATOR-5624](https://openedx.atlassian.net/browse/EDUCATOR-5624)

**What changed?**

- Formerly, a course had to have bulk management enabled and have a master's track. This painted us into a corner where we had to create a special allow-list to enable bulk management on non-master's track courses.
- Instead, rely only on edx-platform to identify when bulk management should/shouldn't be enabled (enabled for **all** master's track courses and special allow-listed based on a course waffle flag).

**⚠️  NOTE** - this must not merge until after the following:
- [x] Allow-listed courses are enabled through course waffle flags in PROD ([AU-35](https://openedx.atlassian.net/browse/AU-35)).
- [x] https://github.com/edx/edx-platform/pull/28080 is merged.
- [x] Waffle flag state is updated (from universal ON to UNSET).

**Developer Checklist**
- [x] Test suites passing
- [x] Documentation and test plan updated, if applicable

**Testing Instructions**

Verify that:
- Visiting a course without a master's track and without the `grades.bulk_management` CourseWaffleFlag "Enabled" does not show bulk management functionality in Gradebook.
- Visiting a course with a master's track shows bulk management functionality in Gradebook.
- Visiting a course with the `grades.bulk_management` CourseWaffleFlag "Enabled" from Django Admin shows bulk management functionality in Gradebook.

**Reviewer Checklist**

- [ ] I've done a visual code review
- [ ] I've tested the new functionality

FYI: @edx/masters-devs-gta


src/data/selectors/index.js
```diff
@@ -221,15 +220,11 @@ export const shouldShowSpinner = (state) => (
...
- * and the course has a masters track.
+ * Returns true iff the course has bulk management enabled
  * @param {object} state - redux state
  * @return {bool} - should show bulk management controls?
  */
-export const showBulkManagement = (state) => (
-  special.hasSpecialBulkManagementAccess(app.courseId(state))
-  || (tracks.stateHasMastersTrack(state) && state.config.bulkManagementAvailable)
-);
+export const showBulkManagement = (state) => (state.config.bulkManagementAvailable);
```
jansenk, line 227: Should we be using a simpleSelector?

jansenk: wait, this *is* a selector, never mind

jansenk (approved):
This looks exactly like what I think we want. Are you going to make LMS changes as a part of this ticket as well?

src/data/selectors/index.js
muselesscreator: eeeh.  If there were more in the config module maybe.  SimpleSelectors are mostly helpful for when there are a number of key fields you want to expose on a model.  For a singleton, is more overhead than necessary.

Honestly, it may just make sense to wrap the config reducer in with the App reducer and then bundle this selector, but this is totally reasonable for now.

leangseu-edx (approved):
LGTM 👍

PR #247, opened 2022-06-16
Pin Npm to 8.5.x to resolve Peer Dependency issues

abdullahwaheed:
### What has changed
Added specific npm package in CI rule to fix peerDependency issue


aht007 (approved):
LGTM

PR #251, opened 2022-07-07
fix: update and resolve dependencies

leangseu-edx:
**TL;DR -** update dependency enough that `npm i` doesn't require `--force`


mattcarter (approved):
approved

PR #273, opened 2022-10-05
refactor: use shared browserslist config

Mashal-m:
Ticket:
[Use shared @edx/browserslist-config](https://github.com/openedx/frontend-wg/issues/47)

* Removes custom `browserslist` configuration in favor of using a [shared configuration](https://github.com/edx/browserslist-config).
* Removes `is-es5` check in CI since ES5 was only needed for IE 11 support which is dropped. The supported browsers defined by the shared configuration all [natively support ES6](https://caniuse.com/?search=es6).

This change reduces the resultant asset bundle size.


Mashal-m:
Closing this because these changes are merged now [PR](https://github.com/openedx/frontend-app-gradebook/pull/299)

PR #276, opened 2022-10-07
style: improve slash and percent sign i18n

jansenk:
We had a PR recently that made some changes so stuff would display correctly in RTL languages. This pulls those out into helper functions so the code reads a bit cleaner, and adds comments explaining what we're doing and why


leangseu-edx (approved):
🏆  LGTM

PR #286, opened 2022-11-28
fix: use getConfig in order to support runtime configuration

kdmccormick:
## Description

Before, gradebook was reading config from `process.env`
directly, which locked the app into using only
static (build-time) configuration. In order to
enable dynamic (runtime) configuration, we update
gradebook to use frontend-platform's standard
configuration interface: `mergeConfig()` and `getConfig()`.

Bump version from 1.5.0 to 1.6.0.
(I would normally just do a patch release for a fix, but the version
 was hasn't been bumped for a while, so adding in full runtime
 configuration support seemed like it warranted a proper
 minor version bump.)

## References

This is a fix to support of https://github.com/overhangio/tutor-mfe/pull/69, which implements https://github.com/openedx/build-test-release-wg/issues/235, which is necessary for the Dec 9 (!) Olive release.

I also have a [draft PR to backport this to Olive.master](https://github.com/openedx/frontend-app-gradebook/pull/287) queued up. Once this PR merges, we can rebase and then merge the backport.

## Developer Checklist
- [X] Test suites passing
   - ~**Reviewers: I'm looking for help on how to resolve the one failing test case -- see my comment below.**~ Thanks Ghassan!
- [x] Codecov passing
- [x] Documentation and test plan updated, if applicable (N/A)

## Testing Instructions

**Making sure it won't break anything:**: Just start up Gradebook locally with Devstack or Tutor and do a general smoke test.

**Making sure it actually fixes runtime configuration**: Follow [@ghassanmas's tutor-mfe PR test instructions](https://github.com/overhangio/tutor-mfe/pull/69), but change `MFE_GRADEBOOK_MFE_APP["repository"]` to `"https://github.com/kdmccormick/frontend-app-gradebook"`.

## Reviewer Checklist

- [ ] I've done a visual code review
- [ ] I've tested the new functionality

FYI: @openedx/content-aurora


src/data/store.test.js
```diff
@@ -60,17 +60,17 @@ describe('store aggregator module', () => {
...
         });
         it('exports thunk and logger middleware, composed and applied with dev tools', () => {
           expect(createStore().middleware).toEqual(
             composeWithDevTools(applyMiddleware(thunkMiddleware, createLogger())),
           );
         });
         afterEach(() => {
-          configuration.SEGMENT_KEY = key;
+          getConfig().SEGMENT_KEY = key;
         });
```
kdmccormick, line 74: I'm looking for help with this case. I'm not sure how to change the result of `getConfig()` for just one test case.

kdmccormick: Currently it fails with:
```
 FAIL  src/data/store.test.js
  ● store aggregator module › exported store › middleware › if no SEGMENT_KEY › exports thunk and logger middleware, composed and applied with dev tools

    expect(received).toEqual(expected) // deep equality

    - Expected  - 0
    + Received  + 4

      Object {
        "withDevTools": Object {
          "applied": Array [
            "thunkMiddleware",
            "logger",
    +       Object {
    +         "map": "EVENTS MAP",
    +         "model": "Segment",
    +       },
          ],
        },
      }

      66 |         });
      67 |         it('exports thunk and logger middleware, composed and applied with dev tools', () => {
    > 68 |           expect(createStore().middleware).toEqual(
         |                                            ^
      69 |             composeWithDevTools(applyMiddleware(thunkMiddleware, createLogger())),
      70 |           );
      71 |         });

      at Object.<anonymous> (src/data/store.test.js:68:44)

```
I presume this is because `getConfig().SEGMENT_KEY = false` isn't actually setting `SEGMENT_KEY` to `false` within the test body.

src/data/store.test.js
```diff
@@ -60,17 +60,17 @@ describe('store aggregator module', () => {
...
       });
       describe('if no SEGMENT_KEY', () => {
-        const key = configuration.SEGMENT_KEY;
+        const key = getConfig().SEGMENT_KEY;
```
arbrandes, line 63: I would've thought we'd need a `let key;` in the scope above this one for the variable to persist until `afterEach()`, but... this did work before, right? :)

kdmccormick:
Oops, looks like I need to add a test that hits the `initialize(...)`  call in index.jsx. Looking at that now.

kdmccormick:
@arbrandes since you're reviewing, let me know if you think of any way to cover the `initialize(...)` call in index.jsx with a test, or whether we should just ignore coverage on those lines. I checked [codecov](https://app.codecov.io/gh/openedx/frontend-app-profile/blob/master/src/index.jsx) for profile, account, and learning, and none of those apps cover `initialize(...)` with a test.

src/data/store.test.js
```diff
@@ -60,17 +60,17 @@ describe('store aggregator module', () => {
...
         beforeEach(() => {
-          configuration.SEGMENT_KEY = false;
+          getConfig.mockImplementation(() => ({ SEGMENT_KEY: false }));
```
arbrandes, line 65: If this is still failing, you could try doing [what frontend-app-learning does](https://github.com/openedx/frontend-app-learning/blob/master/src/instructor-toolbar/InstructorToolbar.test.jsx#L10-L15).

Alternatively, [what EdxHeader does](https://github.com/openedx/frontend-app-gradebook/blob/master/src/components/EdxHeader/test.jsx#L18) (ignoring for a minute the question of what is edX-specific code still doing here), in this repo.

arbrandes (commented):
Looks good!  Left a couple of comments.  Will test it next.

arbrandes:
@kdmccormick 

> any way to cover the `initialize(...)` call in index.jsx

This is what frontend-app-ora-grading does.  Looks like a reasonable test, except for the fact having each-and-every MFE do the same thing feels kinda redundant, just for coverage. 🤷🏼‍♂️

src/data/store.test.js
kdmccormick: It did work! My understanding is that the anonymous function in `afterEach(...)` creates a closure that captures `key`, ensuring that the value of `key` will be accessible as long as the anonymous function exists.

kdmccormick:
@arbrandes Hold on, gradebook is testing the same thing: https://github.com/kdmccormick/frontend-app-gradebook/blob/master/src/index.test.jsx#L49-L57. And that test is definitely running (I know, because I had to fix it :)

Now to figure out why that isn't enough to make Coverage happy...

arbrandes:
I tested the change, and AFAICT it works just fine.

src/data/store.test.js
kdmccormick: @arbrandes I find myself lacking understanding of Jest's mock system, particularly around when patches are enabled and when they are reset. In my head, I'm trying to equate it to Python's mocking framework but that's only getting me so far.

In interest of time, I feel like this is a minor wart that we can leave as-is for now.  Do you agree?

kdmccormick:
I pushed https://github.com/openedx/frontend-app-gradebook/pull/286/commits/c3b16d026b726691caf86ea7b3791032312663a0 , and coverage is now 100%.

src/index.jsx
```diff
@@ -20,6 +21,22 @@ subscribe(APP_READY, () => {
...
+        BASE_URL: process.env.BASE_URL,
+        LMS_BASE_URL: process.env.LMS_BASE_URL,
+        LOGIN_URL: process.env.LOGIN_URL,
+        LOGOUT_URL: process.env.LOGOUT_URL,
+        CSRF_TOKEN_API_PATH: process.env.CSRF_TOKEN_API_PATH,
+        REFRESH_ACCESS_TOKEN_ENDPOINT: process.env.REFRESH_ACCESS_TOKEN_ENDPOINT,
+        DATA_API_BASE_URL: process.env.DATA_API_BASE_URL,
+        SECURE_COOKIES: process.env.NODE_ENV !== 'development',
+        SEGMENT_KEY: process.env.SEGMENT_KEY,
+        ACCESS_TOKEN_COOKIE_NAME: process.env.ACCESS_TOKEN_COOKIE_NAME,
```
kdmccormick, line 36: I have included the config settings that were previously in `config/index.js`. I did **not** pull in everything from `.env`, which contains a lot of unused settings.

I grepped for `process.env` and found two remaining usages:

* A few checks for `process.env.NODE_ENV == 'development'|'production'`. These feel like anti-patterns, but I assume it is OK to leave them as they were, since runtime config won't be overriding `NODE_ENV`.
* `<Footer logo={process.env.LOGO_POWERED_BY_OPEN_EDX_URL_SVG} />` -- This one confuses me, because AFAICT no other MFE has such a setting. Do other MFEs use the "Powered by Open edX" logo? If so, do they configure it some other way? I didn't see it in the default brand package. Anyway, I left this as-is, because tests failed when I tried to put it behind the `getConfig()` interface. Let me know if this seems like an issue Adolfo.

kdmccormick (commented):
@arbrandes Ready for another round of review! I have two questions for you: [this one](https://github.com/openedx/frontend-app-gradebook/pull/286#discussion_r1033692098), and the one posted below.

src/index.jsx
arbrandes: The footer thing does indeed seem like an issue, but not a blocker here.

+1 for not including all the `.env` settings.

+1 for leaving `process.env.NODE_ENV` usages.

src/data/store.test.js
arbrandes: Agreed!  Plus, I didn't realize you had pulled in a fix for the failing test when I wrote that comment.

arbrandes (approved):
LGTM!

PR #289, opened 2022-11-29
feat: ensure lms api synced with latest value in config

ghassanmas:
This change make it possible if LMS url to be changed, that
  the last value will be picked.

  This is simlair openedx/frontend-app-course-authoring/pull/389
  which issue overhangio/tutor-mfe/issues/86, the fixes is needed
  so that dynamic config would work with tutor:
  overhangio/tutor-mfe/pull/69

**TL;DR -** [ A short summary of what this PR does and why ]

- This change is need to so that when LMS URL value is change at runtime. The case of which case happen is when that URL is change through dynamic config. If the LMS url is deifned only at build time  this change shouldn't affect it. 

- The course authroing mfe has simlair issue with CMS url ref: https://github.com/openedx/frontend-app-course-authoring/pull/389 

**What changed?**

- [ More in depth breakdown of changes ]
  - Converting all the use of urls from value to a fucntion, by mkaing the urls obejct return a function than a static object, you will be able to get the latest  value of LMS url.  
  
  Here is a simple example of our case here: 
  
  ### The situation before 
  ```js 
  const LMS  = getConfing().LMS  // assume LMS_1 is example.com
  
  setConfig({LMS: openedx.com})  
  
   LMS === getConfig().LMS  // returs false now its a porblem when any function/ is using LMS_1 varaible
  ```

  ### The situation after 
  ```js 
  const getLMS = ()=> `${getConfing().LMS}`  // assume LMS is example.com
  
  setConfig({LMS: openedx.com})  
  
   getLMS() === getConfig().LMS  // returs true now its not a porblem when any function/ is using LMS_1 varaible 
  ```

 
**Developer Checklist**
- [x] Test suites passing _I tested with tutor_ 
- [x] Documentation and test plan updated, if applicable 
- [ ] Bumped version number [package.json](../package.json)

**Testing Instructions**

- Since this change would fix the following case: 
  - Dynamic API config is enabled by the LMS 
  - The gradebook MFE rely on LMS url to be set at runtime 
    - For this to be `true` then Dynamic config API for this MFE should be relative and a proxy need to be set to forward for the lms, check this PR for more context  overhangio/tutor-mfe/pull/69
    
 - In case dynamic config is off, this change doesn't suppose to affect anything,

**Reviewer Checklist**

FYI: @openedx/content-aurora

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@edx/frontend-app-gradebook",
-  "version": "1.5.0",
+  "version": "1.6.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@edx/frontend-app-gradebook",
-      "version": "1.5.0",
+      "version": "1.6.0",
       "license": "AGPL-3.0",
       "dependencies": {
         "@edx/brand": "npm:@edx/brand-edx.org@^1.3.2",
--- a/src/data/services/lms/api.js
+++ b/src/data/services/lms/api.js
@@ -14,10 +14,10 @@ const { get, post, stringifyUrl } = utils;
 /*********************************************************************************
  * GET Actions
  *********************************************************************************/
-const assignmentTypes = () => get(urls.assignmentTypes);
-const cohorts = () => get(urls.cohorts);
-const roles = () => get(urls.roles);
-const tracks = () => get(urls.tracks);
+const assignmentTypes = () => get(urls.getAssignmentTypesUrl());
+const cohorts = () => get(urls.getCohortsUrl());
+const roles = () => get(urls.getRolesUrl());
+const tracks = () => get(urls.getTracksUrl());
 
 /**
  * fetch.gradebookData(searchText, cohort, track, options)
@@ -45,7 +45,7 @@ const gradebookData = (searchText, cohort, track, options = {}) => {
     [paramKeys.assignmentGradeMax]: options.assignmentGradeMax,
     [paramKeys.assignmentGradeMin]: options.assignmentGradeMin,
   };
-  return get(stringifyUrl(urls.gradebook, queryParams));
+  return get(stringifyUrl(urls.getGradebookUrl(), queryParams));
 };
 
 /**
@@ -53,7 +53,7 @@ const gradebookData = (searchText, cohort, track, options = {}) => {
  * fetches bulk operation history and raises an error if the operation fails
  * @return {Promise} - get response
  */
-const gradeBulkOperationHistory = () => get(urls.bulkHistory)
+const gradeBulkOperationHistory = () => get(urls.getBulkHistoryUrl())
   .then(response => response.data)
   .catch(() => Promise.reject(Error(messages.errors.unhandledResponse)));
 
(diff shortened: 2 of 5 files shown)


arbrandes (commented):
I have yet to test this, but the code looks good.

Why are we updating package-lock in the same commit, though?

ghassanmas:
The change in pacakge-lock.json is realted just to version number which @kdmccormick changed in the previous change in pacakge.json #286. 
So in summary my `npm` updated the pacakge-lock.json to match the version in pacakge.json. anyway I can remove that change if necessary

kdmccormick:
FYI @openedx/content-aurora : The team ping in the PR template didn't work, probably because Ghassan doesn't have direct write access to this repo.

arbrandes:
@muselesscreator, is this something you or somebody on your team could review in the next couple of days?  It's pretty important for the Tutor/Olive release, which is happening next Monday (the 11th) at the latest.

nsprenkle:
@ghassanmas, can you update PR description template fields (lots of stuff left black which should be filled or deleted) and provide testing instructions? Thanks you.

ghassanmas:
@nsprenkle to be able to test this you would need to use enable dynamic config api, plus you would need to rely on the LMS URL, and to do that second you have to set a proxy for the mfe config url... This assuming you don't use tutor. 

Again the thing with testing this,  I am not sure what is your dev env, anyway as @kdmccormick  in simlair [PR](https://github.com/openedx/frontend-app-gradebook/pull/286#issue-1465637411) you can follow the same steps. 


Essetionaly what this change does, is that it enforce that the call of  `getConfig` to be called as from a function not to be called at the file level i.e. [Hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting). It the exact same issue we had with course authoring in this [PR](openedx/frontend-app-course-authoring/pull/389). It's just  in this case it's the LMS url instead of CMS for the course authroing. 

And again to note confuse you @kdmccormick was necessary but it's enough to make an MFE use dyanmic config, an MFE would need to load the config from frontend-platform, and it also needs to ensure that configuration are loaded from a function because otherwise the variabale would be defined once, as mentioned before _Hoisting_

kdmccormick:
@ghassanmas -- Thanks for that info, but please fill it into the PR template.

@nsprenkle -- MFE runtime config will be enabled in Tutor and disabled-by-default for the Devstack/edx.org installation. So:
* If you have time in the next couple of days to provision Tutor and test this change out using runtime config, you could do that. I'm free most of today and would be happy to help you with any provisioning issues.
* Alternatively, you could just test the edx.org static-config use case by running Gradebook normally on devstack (you'd just be confirming that nothing is broken), and then delegate the Tutor runtime-config use case testing to @arbrandes and I.

arbrandes (approved):
I tested this on both a devstack and Tutor.  As expected, there were no changes detectable on the first, and on the latter the MFE now works.

openedx-webhooks:
@ghassanmas 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-app-gradebook&entry.867055334=https://github.com/openedx/frontend-app-gradebook/pull/289&entry.1484655318=https://github.com/ghassanmas&entry.752974735=2022-11-29+08:00&entry.1917517419=2022-12-08+18:21&entry.2133058324=Yes) so we can improve your experience in the future.

PR #299, opened 2023-01-05
Lk/use share browserlist

leangseu-edx:
**TL;DR -**
- start using share browserlist like other repo (Similar to https://github.com/openedx/frontend-app-learning/pull/728)
- upgrade paragon to remove depreciated postcss from bootstrap
- fix grade modal always showing from previous migration https://github.com/openedx/frontend-app-gradebook/pull/275

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,6 +48,7 @@
         "whatwg-fetch": "^2.0.4"
       },
       "devDependencies": {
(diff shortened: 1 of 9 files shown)


nsprenkle (approved):
Builds and fixes the issue for me locally :shipit:

PR #300, opened 2023-01-06
Major version upgrade of paragon

Mashal-m:
Major version upgrade of **@edx/paragon**

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@edx/frontend-component-footer": "^12.0.0",
     "@edx/frontend-component-header": "^4.0.0",
     "@edx/frontend-platform": "^4.2.0",
-    "@edx/paragon": "^19.25.4",
+    "@edx/paragon": "^20.20.0",
     "@edx/reactifex": "^2.1.1",
     "@fortawesome/fontawesome-svg-core": "^1.2.25",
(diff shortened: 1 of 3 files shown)


muselesscreator:
I have no objection to the paragon update, but we don't need react-intl, as we have it imported through frontend-platform

package.json
```diff
@@ -52,6 +52,7 @@
     "react": "16.14.0",
     "react-dom": "16.14.0",
     "react-helmet": "^6.1.0",
+    "react-intl": "^5.25.0",
```
muselesscreator, line 55:
```suggestion
```

Mashal-m: I have uninstalled it.

mattcarter (approved):
Approved

adamstankiewicz:
@Mashal-m @muselesscreator Heads up that there was one additional step needed for this major upgrade to v20 of Paragon. Per the release notes, Paragon v20 now exposes a `paragonMessages` array containing the native translations for Paragon components but the `paragonMessages` are not yet passed to `@edx/frontend-platform` via the `initialize` function in this MFE. 

[inform] I created a GitHub issue about this follow-up task here: https://github.com/openedx/frontend-app-gradebook/issues/340

PR #201, opened 2021-07-01
remove util package as webpack 5 does not like

muselesscreator:
**TL;DR -**
Removes is an out-dated requirement of the StrictDict tool, and is incompatible with Webpack 5.
Alerted to this by David J, who has already had to make this change in the learning MFE

**Developer Checklist**
- [ ] Test suites passing
- [ ] Documentation and test plan updated, if applicable
- [ ] Received code-owner approving review
- [ ] Bumped version number [package.json](../package.json)

**Testing Instructions**

[ How should a reviewer test this PR? ]

**Reviewer Checklist**

Collectively, these should be completed by reviewers of this PR:

- [ ] I've done a visual code review
- [ ] I've tested the new functionality

FYI: @edx/masters-devs-gta

PR #304, opened 2023-01-20
feat: remove util as a dependency

jansenk:
This is an attempt to update and merge an old pr: https://github.com/openedx/frontend-app-gradebook/pull/201

The only remaining change is to remove the `util` package as a dependency. So, here it is.


mattcarter (approved):
approved

PR #321, opened 2023-04-20
feat: add column for full name for masters students

jansenk:
**TL;DR -** Add column for full name

JIRA: [AU-218](https://2u-internal.atlassian.net/browse/AU-218)

FYI: @openedx/content-aurora

### DIFF
--- a/src/components/GradesView/GradebookTable/Fields.jsx
+++ b/src/components/GradesView/GradebookTable/Fields.jsx
@@ -29,16 +29,16 @@ Username.propTypes = {
 };
 
 /**
- * Fields.SimpleText
+ * Fields.Text
  * Simple label field for text value.
  * @param {string} value - value for display
  */
-const SimpleText = ({ value }) => (<span className="wrap-text-in-cell">{value}</span>);
-SimpleText.propTypes = {
+const Text = ({ value }) => (<span className="wrap-text-in-cell">{value}</span>);
+Text.propTypes = {
   value: PropTypes.string.isRequired,
 };
 
 export default StrictDict({
-  SimpleText,
+  Text,
   Username,
 });
--- a/src/components/GradesView/GradebookTable/Fields.jsx
+++ b/src/components/GradesView/GradebookTable/Fields.jsx
@@ -29,16 +29,16 @@ Username.propTypes = {
 };
 
 /**
- * Fields.Email
- * Simple label field for email value.
- * @param {string} email - email for display
+ * Fields.SimpleText
+ * Simple label field for text value.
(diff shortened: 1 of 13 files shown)


jansenk:
Screenshot 2023-04-21 at 12 11 52 PM

src/components/GradesView/GradebookTable/Fields.jsx
```diff
@@ -29,16 +29,16 @@ Username.propTypes = {
...
- * @param {string} email - email for display
+ * Fields.SimpleText
+ * Simple label field for text value.
+ * @param {string} value - value for display
  */
-const Email = ({ email }) => <span className="wrap-text-in-cell">{email}</span>;
-Email.propTypes = {
-  email: PropTypes.string.isRequired,
+const SimpleText = ({ value }) => (<span className="wrap-text-in-cell">{value}</span>);
+SimpleText.propTypes = {
```
nsprenkle, line 37: Nit: is there a reason to call it `SimpleText` and not just `Text`? Not a problem but is more verbose.

src/components/GradesView/GradebookTable/messages.js
```diff
@@ -1,6 +1,11 @@
 import { defineMessages } from '@edx/frontend-platform/i18n';
 
 const messages = defineMessages({
+  fullNameHeading: {
+    id: 'gradebook.GradesView.table.headings.fullName',
+    defaultMessage: 'Full Name*',
```
nsprenkle, line 6: I know this is a pattern we use other places but it is a bad pattern, putting the asterisk styling in the message itself. It should be added as a separate span or other styling.

```suggestion
    defaultMessage: 'Full Name',
```

nsprenkle (commented):
While it wasn't technically specified as an AC, the ticket mentions an additional piece of work that is missing:
- default front-end disabled for large (ie, 100+ learners)

jansenk:
> While it wasn't technically specified as an AC, the ticket mentions an additional piece of work that is missing:
> 
> * default front-end disabled for large (ie, 100+ learners)

Ryan's [most recent comment](https://2u-internal.atlassian.net/browse/AU-218?focusedCommentId=670432) on the ticket says
> hey friends. after a discussion with Mat I think we should replicate the functionality and visibility of the email field for >master-track users. 

>i.e. only show the name data for master-track seats 

This was a day after the text of the ticket was updated. I took this to mean that we were no longer doing the 100 student limit

jansenk:
> I'll check with Ryan

Ryan confirmed that this is the desired behavior

nsprenkle (approved):
LGTM

PR #330, opened 2023-05-08
feat: added "Get Feedback" widget

jszewczulak:
**TL;DR -** "Get Feedback" Widget added to Gradebook

JIRA: [AU-1122](https://2u-internal.atlassian.net/browse/AU-1122)

**What changed?**

- "Get Feedback" widget was added; users can leave feedback on their gradebook experience 

Screenshot 2023-04-25 at 9 25 27 AM

FYI: @openedx/content-aurora


leangseu-edx:
I am not sure if this is intended. If we leave it like that, this will be also included in stage. I think you need to create another `.js` file with the snippet wrapped in `if (process.env.NODE_ENV === 'production')` then import the file in `index.jsx`

jszewczulak:
I see precedent in frontent-app-discussions:

https://github.com/openedx/frontend-app-discussions/blob/84e18b9ed6c5c71fc25a349357980e998e19cb81/public/index.html#L56

https://discussions.stage.edx.org/course-v1:edx+H200+2018/posts

jszewczulak:
Added logic to display widget in PROD only

leangseu-edx (approved):
LGTM

PR #375, opened 2023-12-08
fix: Route with PUBLIC_PATH

arbrandes:
Prior to this change, the MFE would fail to render with any PUBLIC_PATH set.  Because that is now handled entirely by `frontend-platform`, we can avoid referring to PUBLIC_PATH at all.

**Developer Checklist**
- [x] Test suites passing
- [ ] Approving review

**Testing Instructions**

Using Tutor nightly in dev mode, without this patch the gradebook only shows a blank page.  With it, it actually renders.


arbrandes:
This has been both explicitly and implicitly reviewed in Quince, so I'm taking the liberty of merging it in master as well.

